PR #5, opened 2016-07-19
Render prompt on precmd, prefix functions with prompt_geometry

jedahan:
The main fix is to create a setup function that runs prompt_geometry() on precmd. Not sure if its the right way to do things, but its what works for me now.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -112,17 +112,20 @@ prompt_geometry_set_title() {
   prompt_geometry_print_title '%~'
 }
 
-prompt_geometry_setup() {
-  autoload -U add-zsh-hook
-
-  add-zsh-hook preexec prompt_geometry_set_cmd_title
-  add-zsh-hook precmd prompt_geometry_set_title
-
+prompt_geometry_render() {
   PROMPT="
  %(?.$PROMPT_SYMBOL.$EXIT_VALUE_SYMBOL) %{$fg[blue]%}%3~%{$reset_color%} "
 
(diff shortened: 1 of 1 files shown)


jedahan:
Should fix #4

frm:
Thanks for this!

The issue with #4 was using single quotes instead of double quotes, I'm assuming?

jedahan:
The quotes were one issue, but then I found another issue - the prompt was only being rendered one time on startup, not after every command. If you'd like, I can split this pull request into 3 commits that should make things more clear:
- rename private functions to have `prompt_geometry_` prefix
- switch from single to double quote `' => "`
- render prompt on precmd or preexec

frm:
If you could split it for clarity it would be awesome!

geometry.zsh
```diff
@@ -112,17 +112,20 @@ prompt_geometry_set_title() {
...
   PROMPT2=" $RPROMPT_SYMBOL "
   RPROMPT="$(prompt_geometry_git_info)"
 }
 
+prompt_geometry_setup() {
+  autoload -U add-zsh-hook
+
+  add-zsh-hook preexec prompt_geometry_set_cmd_title
+  add-zsh-hook precmd prompt_geometry_set_title
+  add-zsh-hook precmd prompt_geometry_render
```
frm, line 128: Nice change. I noticed the first version was using the `set_title` function to re-render the prompt, which wasn't clear at all, this should be the correct way. 👍

PR #7, opened 2016-07-27
Color triangle based on hostname hash

jedahan:
I thought it would be nice for the color of the triangle to change depending on a hash of the hostname. Makes it easy for me to check at a glance what host I am on.

screen shot 2016-07-27 at 3 13 38 pm


geometry.zsh
```diff
@@ -16,9 +16,10 @@ GIT_UNPUSHED="⇡"
 ZSH_THEME_GIT_TIME_SINCE_COMMIT_SHORT="%{$fg[green]%}"
 ZSH_THEME_GIT_TIME_SINCE_COMMIT_NEUTRAL="%{$fg[white]%}"
 ZSH_THEME_GIT_TIME_SINCE_COMMIT_LONG="%{$fg[red]%}"
+PROMPT_GEOMETRY_COLORED_TRIANGLE=${PROMPT_GEOMETRY_COLORED_TRIANGLE:-false}
 
 prompt_geometry_git_time_since_commit() {
-  if [[ $(git log 2>&1 > /dev/null | grep -c "^fatal: bad default revision") == 0 ]]; then
+  if [[ $(git log -1 2>&1 > /dev/null | grep -c "^fatal: bad default revision") == 0 ]]; then
```
frm: 👍

geometry.zsh
```diff
@@ -112,9 +117,29 @@ prompt_geometry_set_title() {
...
+  local sum=0
+  primes=(1 2 3 5 7 11 13 17 19 23 29 31 37 41 43 47 53 59 61 67 71 73 79 83 89 97 101 103 107 109 113 127 131 137 139 149 151 157 163 167 173 179 181 191)
+  ord_period=$(printf '%d' "'.")
+  for i in {1..${#1}}; do
+    letter=${1[$i]}
+    ord=$(printf '%d' "'$letter")
+    sum=$(($sum + $primes[$i] * $primes[1 + $(($ord - $ord_period))]))
+  done
+  echo $sum
+}
```
frm: I'm kinda conflicted on the hash function but I don't really think we'll need anything better than this.

Why are you using `$ord_period` in the calculation?

frm:
I really like the idea, but is it possible to change it so it doesn't display red? I'd rather add a feature to display in case the user is `root`.

jedahan:
I'll work on simplifying the hash and filtering out lower and upper colors

frm:
Thanks! I just pushed a rather big refactor of the code so it's easier to customize. It might simplify your work as well.

jedahan:
Much easier to customize. I simplified the hashing and color options, how does this look?

frm:
Seems cool! Thanks! Merging.

PR #8, opened 2016-08-02
Allow git time since last commit to be disabled

jedahan:
In an ideal world, we would be able to speed up the git commands to be imperceptible.
In a pragmatic world, we would use async to show the git info.
But today, we can at least disable because there is a 4 second delay on each prompt draw for one of my repositories.


frm:
:+1: thanks. I just want to add a couple more small features before thinking about running zsh-async on it.

PR #10, opened 2016-08-09
Make colorized_symbol screenshot quartersize

jedahan:
It was a bit huge on github pages


frm:
Hi, sorry for taking so long.

I'll accept the PR since the image really is too big, but I'll soon also probably change the documentation and screenshots to showcase more features.

If you have any suggestions regarding that, I'd really appreciate.

Thanks.

PR #12, opened 2016-09-03
Optimize grep command check

SephVelut:
`which` is expensive and `command` is cheaper and POSIX compliant.


frm:
Hi. Thanks for this. Always love those small optimizations!

PR #15, opened 2016-09-19
Fix `bad math expression: empty string` error (OSX)

MrSaints:
A duplicate `for` declaration results in the aforementioned error,
particularly on OSX. `printf` will throw an error due to $i
being empty for the last index.

---

I'm unsure if the duplication is intentional or if it is a mere typo. Nevertheless, this resolves the described error for OSX El Capitan (10.11.5). It is considered a syntax error in bash.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -182,7 +182,7 @@ prompt_geometry_hash_color() {
   fi
 
   local sum=0
-  for for i in {0..${#1}}; do
+  for i in {0..${#1}}; do
     ord=$(printf '%d' "'${1[$i]}")
     sum=$(($sum + $ord))
   done


frm:
Hi! Thanks for this, it was a typo indeed.

Merging.

Just to stay on record in case anyone comes here with the same issue in the future: can you just describe how the issue happened to you? Was it always present when you enabled the prompt symbol color hashing? Also, for the same reason, could you provide the zsh and terminal emulator version?

MrSaints:
Hi @frmendes,

I'm normally on a Lubuntu, and I started using this theme about a week or so ago, with a few modifications. Namely, the `$PROMPT_GEOMETRY_COLORIZE_SYMBOL` was enabled. It was perfectly _fine_ on that machine (appearance wise at least...)

When I sync'd my dotfiles with my OSX machine, it started giving the aforementioned error upon starting my terminal up. There were two errors. One is the `bad math expression` described above. I can't remember the other one. They both pointed to `prompt_geometry_hash_color` as the source of the error. And that is when I noticed the double `for`. With the colorize symbol disabled, the error does not show up.

I plugged the code in https://www.shellcheck.net/ just to validate.

Also, as requested:
- zsh 5.0.8 (x86_64-apple-darwin15.0)
- iTerm 2 Build 3.0.10

Let me know if you have any additional questions :) Happy to help. Thanks!

PR #20, opened 2016-10-14
Added support for virtualenv name in prompt

LuRsT:
This is how it looks like when it's enabled: 

I chose that color for no reason in particular, seemed appropriate.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -18,6 +18,7 @@ GEOMETRY_COLOR_GIT_TIME_SINCE_COMMIT_SHORT=${GEOMETRY_COLOR_GIT_TIME_SINCE_COMMI
 GEOMETRY_COLOR_GIT_TIME_SINCE_COMMIT_NEUTRAL=${GEOMETRY_COLOR_GIT_TIME_SINCE_COMMIT_NEUTRAL:-white}
 GEOMETRY_COLOR_GIT_TIME_SINCE_COMMIT_LONG=${GEOMETRY_COLOR_GIT_TIME_SINCE_COMMIT_LONG:-red}
 GEOMETRY_COLOR_EXIT_VALUE=${GEOMETRY_COLOR_EXIT_VALUE:-magenta}
+GEOMETRY_COLOR_VIRTUALENV=${GEOMETRY_COLOR_PROMPT:-48}
 GEOMETRY_COLOR_PROMPT=${GEOMETRY_COLOR_PROMPT:-white}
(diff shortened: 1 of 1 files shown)


frm:
Hey! Thanks for the PR!

Just a couple of things:
- What do you think about changing the display to the right side of the prompt? Since we're putting most of the similar information on that end, it would be more consistent.
- Can you change the color to something like `green`? It would be more consistent with the rest of the theme regardless of what colorscheme the users activate.

LuRsT:
Sounds sensible, I changed to normal green, replaced the square brackets with normal parenthesis (since it's the default way it appears in bash without any plugin) and moved it to the right. I get a bit of too much spacing when there's no git repo, but I'm not too worried about it.

preview

geometry.zsh
```diff
@@ -217,10 +226,10 @@ prompt_geometry_render() {
   fi
 
   PROMPT="
- %${#PROMPT_SYMBOL}{%(?.$GEOMETRY_PROMPT.$GEOMETRY_EXIT_VALUE)%} %F{$GEOMETRY_COLOR_DIR}%3~%f "
+%${#PROMPT_SYMBOL}{%(?.$GEOMETRY_PROMPT.$GEOMETRY_EXIT_VALUE)%} %F{$GEOMETRY_COLOR_DIR}%3~%f "
```
frm: Can you add this extra space again?

geometry.zsh
```diff
@@ -217,10 +226,10 @@ prompt_geometry_render() {
...
 
   PROMPT2=" $GEOMETRY_SYMBOL_RPROMPT "
-  RPROMPT="$(prompt_geometry_git_info)%f"
+  RPROMPT="$(prompt_geometry_virtualenv) $(prompt_geometry_git_info)%f"
```
frm: Can you move the space between both calls to the `prompt_geometry_virtualenv` `echo` call? The way you put it, it will display that extra space even if users have virtualenv disabled.

frm (changes requested):

LuRsT:
Done!

frm (approved):

frm:
Perfect! Thanks!

glhrmv:
Is this feature still working? I added the following line after `antigen theme geometry-zsh/geometry`:
```
GEOMETRY_RPROMPT+=(geometry_virtualenv)
```
And nothing happens when I `source bin/activate` a virtualenv.

PR #21, opened 2016-10-16
Show human-readable timestamp for long-running commands

jedahan:
screen shot 2016-10-16 at 1 43 56 pm
This lets you see long running commands in rprompt, like in pure.

PR #22, opened 2016-10-17
Show human-readable timestamp for long-running commands

jedahan:
Second try of #21.

There is a weird coloring issue - maybe we should have different colors for git time vs long running time

long running -> white -> yellow -> red
git time -> green -> white -> red

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ What it does:
 
 - display current git branch
 - display colored time since last commit (optional)
+- display the running time of long running commands
 - display state (clean/dirty) of the repo
 - display arrows indicating if you need to pull, push or if you are mid-rebase
 - set the terminal title to current command and directory
@@ -85,6 +86,11 @@ You can have your prompt symbol change colors when running under the root user.
 To activate this option, just set `PROMPT_GEOMETRY_COLORIZE_ROOT` to true. Symbol and color can be customized by changing the `GEOMETRY_SYMBOL_ROOT` and `GEOMETRY_COLOR_ROOT` variables.
 
 Note that this option overrides the color hashing of your prompt symbol.
+
+#### time elapsed for long-running commands
+
+You can optionally show a time display for long-running commands
+by setting the `PROMPT_GEOMETRY_EXEC_TIME` variable to true.
 
 ### FAQs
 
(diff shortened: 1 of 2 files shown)


geometry.zsh
```diff
@@ -48,41 +48,47 @@ GEOMETRY_PROMPT=$(prompt_geometry_colorize $GEOMETRY_COLOR_PROMPT $GEOMETRY_SYMB
 # Flags
 PROMPT_GEOMETRY_GIT_CONFLICTS=${PROMPT_GEOMETRY_GIT_CONFLICTS:-false}
 PROMPT_GEOMETRY_GIT_TIME=${PROMPT_GEOMETRY_GIT_TIME:-true}
+PROMPT_GEOMETRY_EXEC_TIME=${PROMPT_GEOMETRY_EXEC_TIME:-true}
```
frm: Can we default this to false? It feels like this would be added for users that have longer running commands, a subset of the overall users.

geometry.zsh
```diff
@@ -228,8 +234,16 @@ prompt_geometry_render() {
   PROMPT="
  %${#PROMPT_SYMBOL}{%(?.$GEOMETRY_PROMPT.$GEOMETRY_EXIT_VALUE)%} %F{$GEOMETRY_COLOR_DIR}%3~%f "
 
-  PROMPT2=" $GEOMETRY_SYMBOL_RPROMPT "
-  RPROMPT="$(prompt_geometry_virtualenv)$(prompt_geometry_git_info)%f"
```
frm, line 232: Won't this remove the `virtualenv` support?

frm:
Git time is supposed to be green -> white/yellow -> red. I don't have any particular opinion on the colors for the long running commands. Not sure if I understood the issue.

geometry.zsh
jedahan: Fixed, good catch on the mistake.

geometry.zsh
jedahan: Defaulted to false now

jedahan:
The color thing is a non-issue, I was just thinking about it weird

README.md
```diff
@@ -86,6 +87,11 @@ To activate this option, just set `PROMPT_GEOMETRY_COLORIZE_ROOT` to true. Symbo
 
 Note that this option overrides the color hashing of your prompt symbol.
 
+#### time elapsed for long-running commands
+
+You can optionally hide the time display for long-running commands
+by setting the `PROMPT_GEOMETRY_EXEC_TIME` variable to false.
```
frm: This already defaults to false, now. Just this final update before merging.

frm (changes requested):

jedahan:
Now this is super weird, I am getting "normal" as the string display for time elapsed. gonna see what the deal is.

jedahan:
OK fixed a few bugs: the :: separator not working as described, and the actual display. Should be ready to merge now

frm (approved):

frm:
Just merged! Thanks again!

PR #29, opened 2016-11-10
Fix issue with zsh 5.0.x with local variables and subshells

### LINKED ISSUE #25: prompt_geometry_render:local:13: not valid in this context: %F{242}master%f
Alright, so i'm running the latest version of oh-my-zsh and I've gone through the tutorial on your page and keep getting the same error for the theme "geometry":

`prompt_geometry_render:local:13: not valid in this context: %F{242}master%f`

but the triangle will still load...
Any thoughts?

desyncr:
Tested against `5.0.0` and `5.1`. I don't know the root cause of the problem. I should probably checkout zsh-users list or ask on #zsh channel.

Fixes #25


frm:
Perfect! Thanks!

PR #35, opened 2016-11-20
Fix git conflict check

### LINKED ISSUE #17: git_conflicts issue since zplug 2.3.1
Hi,

Updated to zplug `2.3.1`, and since then have had lots of issues with zsh, there is also this issue, where the theme is not displayed, but this error keeps appearing:

`prompt_geometry_git_conflicts:[:10: integer expression expected: 1\n1`

I have an issue posted (here, for some more information)[https://github.com/zplug/zplug/issues/268]

Do you know if this is anything caused on your end, or a symptom of a larger issue?

frm:
I was able to reproduce it so I took the time to fix #17.

`grep` and `ag` were acting out when the list of conflicting files spanned across multiple lines.

In addition to fixing that, I also changed the checks to update the number of files with conflicts. Previously, it used the total number of conflicting files provided by git. This meant that if a file was resolved, it would still show up on the prompt count. Now, if there are no more conflicts in that same file, it won't be added to that count.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -60,7 +60,7 @@ PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT=${PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT:-t
 GEOMETRY_ASYNC_PROMPT_TMP_FILENAME=${GEOMETRY_ASYNC_PROMPT_TMP_FILENAME:-/tmp/geometry-prompt-git-info-}
 
 # Use ag if possible
-GREP=$(command -v ag >/dev/null 2>&1 && echo "ag" || echo "grep")
+GEOMETRY_GREP=$(command -v ag >/dev/null 2>&1 && echo "ag" || echo "grep")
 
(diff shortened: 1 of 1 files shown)


frm:
The most recent push prepares for adding `ripgrep` support. It unfortunately doesn't have the `-o` flag, so I had to come up with a different solution.

tomspeak:
Great work, appreciate it @frmendes

desyncr:
Over all it works. I'm wondering if there is a simplier way to handle conflicting files count. Currently looking at [ls-status](https://git-scm.com/docs/git-status)'s `--short` and `--porcelain`.

frm:
I think using `git-diff` is the best option. I might be wrong but `--name-only` doesn't do very complex calculations for diffing. `--diff-filter` also seems intuitive. The most confusing bit should be the total conflict count but that is related to #36. `ripgrep` doesn't support simpler flags like `-o` so that was the workaround I found.

desyncr:
I went ahead and corrected backtick's usage.

Everything's looking good. Further improvements will be added later.

PR #36, opened 2016-11-20
Adds support for ripgrep

frm:
This PR breaks the git conflict check when there are multiple files with conflicts. See #17.

To avoid breaking changes, this should only be merged after #35.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -64,8 +64,14 @@ GEOMETRY_GIT_NO_COMMITS_MESSAGE=${GEOMETRY_GIT_NO_COMMITS_MESSAGE:-"no commits"}
 GEOMETRY_PROMPT_PREFIX=${GEOMETRY_PROMPT_PREFIX:-"
 "}
 
-# Use ag if possible
-GEOMETRY_GREP=$(command -v ag >/dev/null 2>&1 && echo "ag" || echo "grep")
+# Choose best version of grep
+prompt_geometry_set_grep() {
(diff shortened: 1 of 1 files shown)


geometry.zsh
```diff
@@ -59,8 +59,14 @@ PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT=${PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT:-t
 # Misc configurations
 GEOMETRY_ASYNC_PROMPT_TMP_FILENAME=${GEOMETRY_ASYNC_PROMPT_TMP_FILENAME:-/tmp/geometry-prompt-git-info-}
 
-# Use ag if possible
-GREP=$(command -v ag >/dev/null 2>&1 && echo "ag" || echo "grep")
+# Choose best version of grep
+prompt_geometry_set_grep() {
+  (command -v rg >/dev/null 2>&1 && echo "rg") \
```
desyncr, line 69: What do you think about this?

    (( $+commands[rg] )) && echo "rg" || (( $+commands[ag] )) && echo "ag" || echo "grep"

I have tested it against `command -v` and it's 10x times faster.

    # command -v
    zsh ./test-command.zsh  0.36s user 0.22s system 106% cpu 0.545 total

    # $+commands
    zsh ./test-command-2.zsh  0.02s user 0.02s system 96% cpu 0.042 total

desyncr (approved):

desyncr:
Looking good as is. I'm going ahead and merging this as well.

PR #39, opened 2016-11-23
Ensure multiple concurrent async process are not run

desyncr:
Kill process and subprocess launched in async function.

This issue could cause to have multiple `prompt_geometry_setup_async_prompt` functions running in parallel if the `git_info` prompt is slow enough or the user hits `enter` key in a fast sequence.

- [x] Test on BSD/GNU (Mac, FreeBSD, Linux variants)

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -306,7 +306,7 @@ prompt_geometry_setup_async_prompt() {
 
         # kill child if necessary
         if [[ "${GEOMETRY_ASYNC_PROMPT_PROC}" != 0 ]]; then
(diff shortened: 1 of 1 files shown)


frm:
Does this solve the slow prompt when pressing but not releasing the `enter` key?

frm (approved):

desyncr:
It does so to some extent. Before it could even freeze. Now the delay is less and I haven't seen it freeze.

frm:
Great! Nice catch!

PR #43, opened 2016-11-29
Initial plugin architecture set up

### LINKED ISSUE #42: RFC: Plugin architecture
There are lots of tools that can be added to the prompt, they are useful for some but not generally useful for everyone, naturally. This causes the code to grow large and maintenance becomes a problem. This will become an issue as more prompts are added.

To be able to manage this issue I propose to turn most informative prompts (git, virtualenv, python, etc) into a separate file that can be loaded at runtime. This will keep separate the base prompt from the informative prompts, making maintenance a bit easier. Also performance-wise seems reasonable to keep unloaded unused prompts.

This structure will also provide the ability to create custom -user made- prompts, outside the geometry code base.

Implementation will require to establish a common interface for the prompts (function and variable names) as well as a globally available array to register a given plugins/prompt. Similar to https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md

Thoughts?

Edit: Probably Fixes #24

desyncr:
- Moved most helper functions (`prompt_geometry_colorize`, `prompt_geometry_hash_color`, `prompt_geometry_seconds_to_human_time`, `prompt_geometry_setup_async_prompt`) to `lib/` directory.

- Created `plugins/` directory with default plugins: git, exec_time

- Move most variables/flags to it's plugin file

- It does support custom prompts! [Example](https://gist.github.com/desyncr/6262aa698ecbcf773378cf1ea14d0e2c).

- It does works :D

## TODO

  - [x] Unregister/shutdown functions
  - [x] Avoid registering a plugin multiple times
  - [ ] Enforce code/naming conventions
  - [ ] Update `README.md`
  - [ ] Review backward compatibility

Fixes #42


geometry.zsh
```diff
@@ -2,263 +2,40 @@
...
 PROMPT_GEOMETRY_COLORIZE_SYMBOL=${PROMPT_GEOMETRY_COLORIZE_SYMBOL:-false}
 PROMPT_GEOMETRY_COLORIZE_ROOT=${PROMPT_GEOMETRY_COLORIZE_ROOT:-false}
-PROMPT_VIRTUALENV_ENABLED=${PROMPT_VIRTUALENV_ENABLED:-false}
-PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME=${PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME:-5}
-PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT=${PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT:-true}
-PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY=${PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY:-true}
-PROMPT_DOCKER_MACHINE_ENABLED=${PROMPT_DOCKER_MACHINE_ENABLED:-false}
 PROMPT_GEOMETRY_SHOW_RPROMPT=${PROMPT_GEOMETRY_SHOW_RPROMPT:-true}
+PROMPT_GEOMETRY_RPROMPT_ASYNC=${PROMPT_GEOMETRY_RPROMPT_ASYNC:-true}
+PROMPT_GEOMETRY_ENABLE_PLUGINS=${PROMPT_GEOMETRY_ENABLE_PLUGINS:-true}
```
desyncr, line 33: This configuration is probably unnecessary, having `PROMPT_GEOMETRY_SHOW_RPROMPT`.

lib/async.zsh
```diff
@@ -0,0 +1,33 @@
+# Misc configurations
+GEOMETRY_ASYNC_PROMPT_TMP_FILENAME=${GEOMETRY_ASYNC_PROMPT_TMP_FILENAME:-/tmp/geometry-prompt-git-info-}
```
desyncr: Change default tmp file name.

lib/async.zsh
```diff
@@ -0,0 +1,33 @@
...
+        fi
+
+        -prompt-geometry-async-function &!
+        GEOMETRY_ASYNC_PROMPT_PROC=$!
+    }
+    add-zsh-hook precmd prompt_geometry_precmd
+
+    TRAPUSR1() {
+        # read from temp file
+        RPROMPT="$(<${GEOMETRY_ASYNC_PROMPT_TMP_FILENAME}$$)"
```
desyncr: Remove tmp file.

lib/plugin.zsh
```diff
@@ -0,0 +1,45 @@
...
+fi
+
+# List of active plugins
+typeset -ga GEOMETRY_PROMPT_PLUGINS
+
+# Set up default plugins
+geometry_plugin_setup() {
+  for plugin in $GEOMETRY_PROMPT_DEFAULT_PROMPTS; do
+      source "$GEOMETRY_ROOT/plugins/$plugin.zsh"
+      geometry_plugin_register $plugin
```
desyncr: Probably better to leave plugin to self-register

lib/plugin.zsh
```diff
@@ -0,0 +1,45 @@
...
+  done
+}
+
+# Register a plugin
+geometry_plugin_register() {
+  local plugin=$1
+  if [[ $plugin == "" ]]; then
```
desyncr: Use argument count

lib/plugin.zsh
```diff
@@ -0,0 +1,45 @@
...
+
+# Renders the registered plugins
+geometry_plugin_render() {
+    local rprompt=""
+    local render=""
+
+    for plugin in $GEOMETRY_PROMPT_PLUGINS; do
+        render=$(geometry_prompt_${plugin}_render)
+        if [[ $render != "" ]]; then
+            rprompt+="$render "
```
desyncr: Add configuration for plugin separator character(s).

lib/time.zsh
```diff
@@ -0,0 +1,32 @@
...
+prompt_geometry_seconds_to_human_time() {
+  local color=""
+  local human="" total_seconds=$1
+  local days=$(( total_seconds / 60 / 60 / 24 ))
+  local hours=$(( total_seconds / 60 / 60 % 24 ))
+  local minutes=$(( total_seconds / 60 % 60 ))
+  local seconds=$(( total_seconds % 60 ))
+
+  if $PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT; then
+    if (( days > 0 )); then
```
desyncr: Move each if branch into it's own functions.

plugins/README.md
```diff
@@ -0,0 +1,5 @@
+# Geometry plugins
```
desyncr, line 1: - Add example plugin structure
- Add `setup` return values definition
- Add `render` return values definition

geometry.zsh
```diff
@@ -2,263 +2,40 @@
...
-# Define how to colorize before the variables
-prompt_geometry_colorize() {
-  echo "%F{$1}$2%f"
-}
+GEOMETRY_ROOT=${0:A:h}
+source "$GEOMETRY_ROOT/lib/async.zsh"
+source "$GEOMETRY_ROOT/lib/plugin.zsh"
+source "$GEOMETRY_ROOT/lib/time.zsh"
+source "$GEOMETRY_ROOT/lib/color.zsh"
+source "$GEOMETRY_ROOT/lib/grep.zsh"
```
desyncr, line 10: Not sure about this. Should we have a compilation script to bundle everything into geometry?

lib/async.zsh
```diff
@@ -0,0 +1,33 @@
+# Misc configurations
+GEOMETRY_ASYNC_PROMPT_TMP_FILENAME=${GEOMETRY_ASYNC_PROMPT_TMP_FILENAME:-/tmp/geometry-prompt-git-info-}
+
+GEOMETRY_ASYNC_PROMPT_PROC=0
+prompt_geometry_setup_async_prompt() {
```
desyncr: Rename to `geometry_async_setup`?

lib/color.zsh
```diff
@@ -0,0 +1,20 @@
+# Define how to colorize before the variables
+prompt_geometry_colorize() {
```
desyncr, line 2: Rename to `geometry_color_colorize`?

plugins/README.md
frm: I have a README almost ready, was working for the current version. Just need to adapt it. Can you change the base branch to something like `2.0` instead of `master`? That way I can contribute as well.

lib/async.zsh
frm: 👍  agreed

geometry.zsh
frm: What is the first variable supposed to do?

lib/async.zsh
frm: Do you have any ideas regarding an alternative for tmp files? I can see piping doing the job but is it that much necessary?

lib/plugin.zsh
frm: I'd prefer having a plugin self-register.

lib/color.zsh
frm: Not sure. `color_colorize` doesn't read well to me. Since we keep using this function throughout the prompt why not a shorthand name? Even if it's `-g-color`, just to save characters.

geometry.zsh
frm: I'd rather have that. And have it make sure we only source it once.

frm:
Really liking the way this is going. Great job!

I'd recommend, like I said in another comment, putting the PR on a different branch. That way I and someone who would like to help can jump in easily.

A lot of comments you had there could be in a TODO list, together with the list already on the PR description. I'd like to have some issues and improvements flagged for people looking for their first PR. Props for flagging some issues, really helps with that.

geometry.zsh
desyncr: `PROMPT_GEOMETRY_SHOW_RPROMPT` is used to activate or deactive the rprompt, meanwhile `PROMPT_GEOMETRY_ENABLE_PLUGINS` activates the plugins (internal and custom). The issue is -currently- the rprompt only shows plugins.

lib/plugin.zsh
desyncr: :+1:

plugins/README.md
desyncr: Alright, I'm gonna create `v2` branch and target this PR to it.

lib/color.zsh
desyncr: `geometry_color_colorize` would follow the format `geometry_<lib>_<function>`. I could also add, as an alias, a `-g-color` function.

geometry.zsh
desyncr: Do you mean to use `source` instead of a makefile-like approach?

lib/async.zsh
desyncr: Well, there is [zsh-async](https://github.com/mafredri/zsh-async).

plugins/README.md
desyncr: Done. It was created from `master`. I guess we can already merge this feature branch and continue working on `v2`.

lib/time.zsh
```diff
@@ -54,12 +54,15 @@ typeset -g geometry_time_color
...
+  local long_format=$2
+
   local days=$(( total_seconds / 60 / 60 / 24 ))
   local hours=$(( total_seconds / 60 / 60 % 24 ))
   local minutes=$(( total_seconds / 60 % 60 ))
   local seconds=$(( total_seconds % 60 ))
 
-  if $PROMPT_GEOMETRY_GIT_TIME_SHORT_FORMAT; then
+  # It looks redundant but it seems it's not
+  if [[ $long_format == true ]]; then
```
frm, line 65: Do we need `== true` ?

lib/color.zsh
frm: Second option seems nice.

geometry.zsh
frm: Oh, I thought you meant sourcing them all. A Makefile-like approach would have to be on install, wouldn't it?

frm (approved):

lib/time.zsh
desyncr: Not sure tbh. I tried `[[ $long_format ]]` and it messed up. Gonna take a look later today to see what was going on.

geometry.zsh
desyncr: On a build step. Maybe it's too cumbersome as of now. I'm gonna leave it as it is, sourcing them as currently implemented.

desyncr:
Gonna merge to `develop`. I'm not squashing commits as of now to keep the changelog.

All development regarding plugins should be done in a feature branch branched off `v2`.

Issue #40, opened 2016-11-25
Geometry compatibility with zsh 5.0.x


desyncr:
Currently geometry won't run on `zsh 5.0.x` due to a simple syntax issue present in a couple of places, ej. https://github.com/frmendes/geometry/blob/master/geometry.zsh#L124, https://github.com/frmendes/geometry/blob/master/geometry.zsh#L223.

Error message:

    geometry.plugin.zsh:124: parse error near `]]'

These are easy to fix issues for anyone interested in contributing.

desyncr:
Fixed in #46

PR #46, opened 2016-12-03
Fix geometry compatibility with zsh 5.0.x [#40]

mihaiolteanu:
Use the explicit form to test whether the conditional expression
expands to a string of non-zero length.


desyncr:
Looking good! Thank you! :+1:

PR #47, opened 2016-12-05
load zsh/datetime module for EPOCHSECONDS support

jedahan:
Might want to have an `|| { #disable prompt_geometry_exec_time }` in the future, but I needed this for EPOCHSECONDS to work


desyncr:
Thank you for looking into this. 👍

PR #48, opened 2016-12-08
Change plugin location

frm:
As I was rewriting the README, I realized it's much cleaner to have each plugin write its own documentation. However, that's not possible if all plugins are under the same directory.

To solve that I moved them to `plugins/<plugin_name>/` and changed the plugin file to `plugin.zsh`.

What's your opinion @desyncr?


desyncr:
@frmendes Looks good to me. This way we can have a `plugins/README.md` to describe plugin structure and general info, examples etc. and then a readme for each plugin separately. :+1:

frm:
That was the plan. PR on the README coming probably tomorrow.

desyncr (commented):
Looking good!

PR #64, opened 2016-12-14
Removed enabled tags in two plugins

keelan542:
Not really sure if this is the fix for the issue as this is my first contribution. Removed enabled flags for two plugins.


frm:
Not quite. You're almost there though! You see, if you load the theme, you can try to enable these two plugins by setting `GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine)` in your `~/.zshrc` file, before loading geometry.

If you do try to it like that, however you will notice they won't be enabled. This is because you still need to set `PROMPT_DOCKER_MACHINE_ENABLED` and `PROMPT_VIRTUALENV_ENABLED` to `true`.

What we want here is not having to do that at all. We want to active the behaviour just by setting `GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine)`. Those variables don't seem to actually needed at all!

keelan542:
Okay, so do I just remove the `true` values or do I also remove the `false` values ? Thank you

frm:
You can just delete the variables and remove any reference to them. The `setup` functions don't even need to return anything, they can just be empty: `geometry_prompt_virtualenv_setup() {}`

frm:
That's it! Just one final request: in order to keep the code clean and consistent throughout the codebase, can you put the empty `setup` functions in a single line? `geometry_prompt_virtualenv_setup() {}`

keelan542:
Done ! Thanks for your help :)

frm:
Perfect! Squashing and merging it. Congrats on your first contribution! 🎉 🎉 🎉

keelan542:
Thank you !

desyncr:
Late to the party. Congrats @keelan542 for your first open source contribution! The first of many more I hope!

PR #67, opened 2016-12-17
Handle aliased `rm` command and tmp files

### LINKED ISSUE #59: remove /tmp/geometry-rprompt-info-69286?
After every prompt, I'm getting the following, which can be dismissed by pressing enter:
```
remove /tmp/geometry-rprompt-info-69286?
```

desyncr:
Fixes #59. 

This PR only handles that particular issue. Moved functionality to clean tmp files to the setup phrase rather than to do it on every cycle.

### DIFF
--- a/lib/async.zsh
+++ b/lib/async.zsh
@@ -1,11 +1,13 @@
 # Misc configurations
-typeset -g GEOMETRY_ASYNC_TMP_FILENAME=${GEOMETRY_ASYNC_TMP_FILENAME:-/tmp/geometry-rprompt-info-}
+typeset -g GEOMETRY_ASYNC_TMP_PATH=${GEOMETRY_ASYNC_TMP_PATH:-/tmp}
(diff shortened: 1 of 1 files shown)


frm (approved):
This seems like a simple, straightforward option. 👍

desyncr:
One last thing that was kind of concerning me is that what if the user mess up with `GEOMETRY_ASYNC_TMP_FILENAME`? It may break the system or lost data.

I think I'm gonna refactor this up by renaming this configuration variable to `GEOMETRY_ASYNC_TMP_PATH` in order to:

- Let the user configure the path there to store tmp files
- Use a fixed prefix for file names

So I'll finaly use `\rm "$GEOMETRY_ASYNC_TMP_PATH/.geometry-rprompt-info-*"`.

This will allow me to be sure what I'm removing (only files what that particular prefix), and protect the user from any miss use of this configuration.

Unless I'm missing something I think this approach is way more secure.

What do you think?

frm:
A bit late, but I agree, yes.

PR #68, opened 2016-12-17
Small performance improvement on large repos

### LINKED ISSUE #60: Lag and high CPU usage on large repositories holding enter-key
On this particular scenario, where the user holds the enter-key and the current directory is a huge repository, geometry (async library, in fact) may consume a lot of CPU and for this reason the prompt may "lag" and seem "slow".

I attribute the high CPU usage to the [exec, kill, and re-exec](https://github.com/frmendes/geometry/blob/master/lib/async.zsh#L17) of git subprocesses to print the git built-in plugin.

We could solve this by performing a small cache (1 sec, half a sec) of rprompt. Any other idea? Suggestion?

desyncr:
▲ /path/to/large-repo time git status --porcelain --ignore-submodules &> /dev/null
    git status --porcelain --ignore-submodules &> /dev/null  0.69s user 0.03s system 99% cpu 0.733 total

    ▲ /path/to/large-repo time git status --porcelain --ignore-submodules HEAD &> /dev/null
    git status --porcelain --ignore-submodules HEAD &> /dev/null  0.20s user 0.00s system 99% cpu 0.206 total

It's a quite noticeable improvement but I don't know if I'm missing/breaking something up.

Partially fixes #60

### DIFF
--- a/plugins/git/plugin.zsh
+++ b/plugins/git/plugin.zsh
@@ -55,7 +55,11 @@ prompt_geometry_git_branch() {
 
 prompt_geometry_git_status() {
   if test -z "$(git status --porcelain --ignore-submodules HEAD)"; then
-    echo $GEOMETRY_GIT_CLEAN
+    if test -z "$(git ls-files --others --modified --exclude-standard)"; then
+      echo $GEOMETRY_GIT_CLEAN
+    else
+      echo $GEOMETRY_GIT_DIRTY
+    fi
   else
     echo $GEOMETRY_GIT_DIRTY
   fi
--- a/plugins/git/plugin.zsh
+++ b/plugins/git/plugin.zsh
@@ -54,7 +54,7 @@ prompt_geometry_git_branch() {
 }
 
 prompt_geometry_git_status() {
-  if test -z "$(git status --porcelain --ignore-submodules)"; then
+  if test -z "$(git status --porcelain --ignore-submodules HEAD)"; then
     echo $GEOMETRY_GIT_CLEAN
   else
     echo $GEOMETRY_GIT_DIRTY


plugins/git/plugin.zsh
```diff
@@ -54,7 +54,7 @@ prompt_geometry_git_branch() {
 }
 
 prompt_geometry_git_status() {
-  if test -z "$(git status --porcelain --ignore-submodules)"; then
+  if test -z "$(git status --porcelain --ignore-submodules HEAD)"; then
```
frm, line 57: The issue here is that using `HEAD` it only checks for tracked files.

Although we can add something specific for untracked files, like `git ls-files --others --exclude-standard` (idk if we can optimize that command further) and use your version for tracked files as an optimization: quickly check if tracked files are dirty, if not take a longer look at untracked files.

Most of the time repos are dirty because of tracked files. Programmers aren't adding new files that often and if they are in a repo which is clean, they're probably going make dirty soon enough. We can go with 2 checks if the total time isn't *much* longer than the current time in the live version. Most of the time, the second check won't even be needed.

Can you measure performance on your end so we can have comparable data from the same sources?

frm (changes requested):

plugins/git/plugin.zsh
desyncr: Sure thing.

desyncr: # Current live
    ▲ time zsh ./test.zsh                  master :: 135d :: ⬡
    zsh ./test.zsh  1.11s user 0.08s system 113% cpu 1.054 total

    # Two step
    ▲ time zsh ./test.zsh                  master :: 135d :: ⬡
    zsh ./test.zsh  0.35s user 0.11s system 102% cpu 0.449 total

frm: Does the second test include untracked files? Or just regular dirty files?

desyncr: # Old
    ▲ time zsh ./test.zsh                                                                               master :: 136d :: ⬡ 
    zsh ./test.zsh  0.99s user 0.09s system 107% cpu 1.007 total

    # two step
    ▲ time zsh ./test.zsh                                                                               master :: 136d :: ⬡ 
    zsh ./test.zsh  0.36s user 0.11s system 102% cpu 0.459 total

frm (approved):

PR #70, opened 2016-12-21
My attempt

cnguy:
Hello!

This creates the desired output.

The problem is that I grepped all the files in the directory and found that the line I deleted was the only place in which $GEOMETRY_PLUGIN_SEPARATOR was used.

`typeset` -g GEOMETRY_PLUGIN_SEPARATOR=${GEOMETRY_PLUGIN_SEPARATOR:-" "}

Without it, the output is correct, but I'm not sure if that's how I'm supposed to approach the problem.

edit: 6:20:34 PM
Reverted that commit and commited a different change.

edit: 636:57 PM
My second commit doesn't address the actual problem as well. Woops... :(

### DIFF
--- a/plugins/git/plugin.zsh
+++ b/plugins/git/plugin.zsh
@@ -54,8 +54,12 @@ prompt_geometry_git_branch() {
 }
 
 prompt_geometry_git_status() {
-  if test -z "$(git status --porcelain --ignore-submodules)"; then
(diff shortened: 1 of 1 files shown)


ChauTNguyen:
So I figured out that the way I was doing it was wrong, and now I did it another way.. However, I deleted the original branch I used to make the first pull request...

PR #71, opened 2016-12-21
Added if statements to fix spacing

cnguy:
Hello again -- sorry for the last pull request. I'm sending a new one after I thought about the code for a while.

Last PR: https://github.com/frmendes/geometry/pull/70

I added if statements since I believe the problem is that in certain cases, there are unnecessary spaces being generated.

--
**L98** `echo "$(prompt_geometry_git_rebase_check) $(prompt_geometry_git_remote_check)"`

1. Even when the functions `prompt_geometry_git_rebase_check()` and `prompt_geometry_git_remote_check()` both do not output anything, a space is generated in between the empty results of the two functions.
Unintended resulting string: `" "`

2. Another check to avoid an unnecessary space has to be made in the case that the first statement fails, but the second statement passes.
Unintended resulting string: `" $(prompt_geometry_git_remote_check)"`

--
**L143** `echo -n "$(prompt_geometry_git_symbol) $(prompt_geometry_git_branch) $conflicts::$time $(prompt_geometry_git_status)"`

3. An unnecessary space occurs after `prompt_geometry_git_symbol()` even if the function did not output anything. I added an if statement to check for that case. If the function does output something, then a space needs to be added.
Unintended resulting result: `" $(prompt_geometry_git_branch) $conflicts::$time $(prompt_geometry_git_status)"`

--- Edit 5:26 to make the post clearer

### DIFF
--- a/plugins/git/plugin.zsh
+++ b/plugins/git/plugin.zsh
@@ -95,7 +95,24 @@ prompt_geometry_git_remote_check() {
 }
 
 prompt_geometry_git_symbol() {
-  echo "$(prompt_geometry_git_rebase_check) $(prompt_geometry_git_remote_check)"
+  local git_rebase="$(prompt_geometry_git_rebase_check)"
+  local git_remote="$(prompt_geometry_git_remote_check)"
+  local render=""
+
+  if [[ $git_rebase != "" ]]; then
+    render+="$git_rebase"
+    
+    if [[ $git_remote != "" ]]; then
+      render+=" $git_remote"
+    fi
+
+  fi
+
+  if [[ $git_remote != "" ]]; then
+    render+="$git_remote"
+  fi
+
+  echo -n $render
 }
 
 prompt_geometry_git_conflicts() {
@@ -140,7 +157,16 @@ geometry_prompt_git_render() {
       fi
     fi
 
-    echo -n "$(prompt_geometry_git_symbol) $(prompt_geometry_git_branch) $conflicts::$time $(prompt_geometry_git_status)"
+    local render="$(prompt_geometry_git_symbol)"
+
+    if [[ $render != "" ]]; then
+      render+=" "
+    fi
+
+    render+="$(prompt_geometry_git_branch) $conflicts::$time $(prompt_geometry_git_status)"
+
+    echo -n $render
+
   fi
 }
 
--- a/plugins/git/plugin.zsh
+++ b/plugins/git/plugin.zsh
@@ -95,20 +95,19 @@ prompt_geometry_git_remote_check() {
 }
 
 prompt_geometry_git_symbol() {
+  local render=""
   local git_rebase="$(prompt_geometry_git_rebase_check)"
   local git_remote="$(prompt_geometry_git_remote_check)"
-  local render=""
 
-  if [[ $git_rebase != "" ]]; then
+  if [[ -n $git_rebase ]]; then
     render+="$git_rebase"
-    
-    if [[ $git_remote != "" ]]; then
-      render+=" $git_remote"
-    fi
-
   fi
 
-  if [[ $git_remote != "" ]]; then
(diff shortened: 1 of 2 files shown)


desyncr:
Looks good to me! May be there are minor code style issues but it does work well. @frmendes What do you think?

By the way thank you @ChauTNguyen !

desyncr (approved):

frm:
Sorry it took me so long. LGTM, it does work well, but I wonder if we can simplify it. I'm up for merging right away and later on working the code style or doing it right away. What do you think @desyncr and @ChauTNguyen ?

desyncr:
@frmendes Well taking a further look it seems there may be an issue in [this](https://github.com/frmendes/geometry/pull/71/files#diff-a24157302c0636b9a78078bd31d9ec7bR102) line. As the logic may show two `remote checks` symbols as shown in the screenshot below.

screenshot from 2016-12-22 19-47-57

The scenario is present while rebasing and having either unpulled or unpushed commits.

@ChauTNguyen Let me know if you are up to make these changes :)

plugins/git/plugin.zsh
```diff
@@ -95,7 +95,24 @@ prompt_geometry_git_remote_check() {
 }
 
 prompt_geometry_git_symbol() {
-  echo "$(prompt_geometry_git_rebase_check) $(prompt_geometry_git_remote_check)"
+  local git_rebase="$(prompt_geometry_git_rebase_check)"
+  local git_remote="$(prompt_geometry_git_remote_check)"
+  local render=""
+
+  if [[ $git_rebase != "" ]]; then
```
desyncr: I'd rather use `-n` instead of `!=`:

    if [[ -n $git_rebase ]]; then

[Here](http://zsh.sourceforge.net/Doc/Release/Conditional-Expressions.html) is the list of conditionals available.

plugins/git/plugin.zsh
```diff
@@ -95,7 +95,24 @@ prompt_geometry_git_remote_check() {
...
+    render+="$git_rebase"
+    
+    if [[ $git_remote != "" ]]; then
```
desyncr: To avoid code duplication I'd rather structure these conditionals in this way (pseudocode):

    if rebase; then
       render+=rebase
    fi

    # if both symbols are present (rebase && remote)
    if rebase && remote; then
      render+=" "
    fi

    if remote; then
       render+=remote
    fi

cnguy:
@desyncr @frmendes Oops, I see the error in the code. I'll try to fix the code and its style.

desyncr (changes requested):

desyncr:
@ChauTNguyen Thank you! Let us know if you need any kind of feedback!

cnguy:
@desyncr Thanks for the list! I had no idea zsh had commands like that; I forgot that it's a scripting language or whatever like bash so it has flags too...

edit: woops ignore that. that was my error..

I meant line 74 of lib/plugin.zsh

`
    if [[ $render != "" ]]; then
         rprompt+="$render$GEOMETRY_PLUGIN_SEPARATOR"
    fi
`

desyncr:
@ChauTNguyen Yes, to keep the code consistent.
There may be other occurrences around the code both don't mind them; we'll review code style and conventions in a future PR.

desyncr:
Regarding...

> I meant line 74 of lib/plugin.zsh
> if [[ $render != "" ]]; then rprompt+="$render$GEOMETRY_PLUGIN_SEPARATOR" fi

Well, if it's the only one yes you may change that as well.

Looks great so far! Once that last change is done I think we can merge this!

Thank you very much for you contribution @ChauTNguyen!

desyncr (approved):
Looks great! Thank you so much!

desyncr:
Merged! Thank you so much @ChauTNguyen ! Great work!

PR #75, opened 2016-12-24
First attempt at fixing exec_time display bug

cnguy:
Hello,

This is my first attempt at fixing the exec_time display bug here: https://github.com/frmendes/geometry/issues/72

### DIFF
--- a/plugins/exec_time/plugin.zsh
+++ b/plugins/exec_time/plugin.zsh
@@ -32,8 +32,20 @@ geometry_prompt_exec_time_setup() {
   return true
 }
 
+check_empty_buffer() {
+  if [[ $#BUFFER == 0 ]]; then
+    current_command_is_empty=true
+  else
+    current_command_is_empty=false
+  fi
+}
+
+zle -N zle-line-finish check_empty_buffer
+
 geometry_prompt_exec_time_render() {
-  echo "$prompt_geometry_command_exec_time"
+  if [ "$current_command_is_empty" = false ]; then
+    echo "$prompt_geometry_command_exec_time"
+  fi
 }
 
 # Self-register plugin


desyncr:
I think it's a good time to take a closer look at how this plugin works. I think it's quite convoluted as of now for such a simple task as to display how long a command took.

What I see as the main problem is that we're not cleaning up `prompt_geometry_command_exec_time` after it's rendered. Same happens with `prompt_geometry_command_timestamp`, both should be cleared out. Ex:

```sh
prompt_geometry_check_command_exec_time() {
  integer elapsed
  # default value for exec_time is an empty string (ie, it won't be rendered), if we don't clear this up
  # it may be rendered each time
  prompt_geometry_command_exec_time=
  (( elapsed = EPOCHSECONDS - ${prompt_geometry_command_timestamp:-$EPOCHSECONDS} ))
  if (( elapsed > $PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME )); then
    prompt_geometry_command_exec_time="$(prompt_geometry_seconds_to_human_time $elapsed)"
  fi
  # Clear timestamp after we're done calculating exec_time
  prompt_geometry_command_timestamp=
}
```

Another thing is that we're hooking `precmd` for `prompt_geometry_set_command_timestamp` but we should rather use `preexec`. Check out the [docs](http://zsh.sourceforge.net/Doc/Release/Functions.html) (ctrl+f, precmd). I'd use the following setup:

```sh
# sets the timestamp before running a command
add-zsh-hook preexec prompt_geometry_set_command_timestamp

# before rendering the prompt do run the checks
add-zsh-hook precmd prompt_geometry_check_command_exec_time
```

It think this way we'll simplify a bit how this plugin work. What do you think? @ChauTNguyen

cnguy:
@desyncr 

Ahh, I see.  Thank you. Yeah, preexec makes a lot of sense now that I've learned a bit about it.

Clearing the values after they're used makes a lot of sense to me as well. I added the changes you typed out, but I couldn't fix the original bug. Am I missing something?

desyncr:
Check out this implementation: https://gist.github.com/desyncr/4573fa6aea5308aa7813e91ad30e9c7c

From here it seems to work.

desyncr:
@ChauTNguyen Do you need any help? Did my suggestion worked at all? Let me know :)

cnguy:
Hello,  @desyncr!

I have tried the code, but I encountered a problem. The elapsed time is always 0 because the program sets the timestamp and then right after subtracts the timestamp from $EPOCHSECONDS (they have the same values then). I think it's because the program loses track of the previous successful command's timestamp.

I've been trying various ways tonight to try to store the old timestamp and keep the hooks and stuffs, but I can't quite get it correct.

Am I misunderstanding something/doing something wrong (I'm not too good at zsh...)?

desyncr:
As I understand the flow should go as follows:

### Case 1: Command

- User submits a command (ex, `sleep 7<enter>`)
- `preexec` hook is handled by `prompt_geometry_set_command_timestamp`
- `prompt_geometry_command_timestamp` is set to `EPOCHSECONDS`
- The command ran and exists after ~7s

Then:

- `precmd` hook is handled by `prompt_geometry_check_command_exec_time`
- We clear up previous values in `prompt_geometry_command_exec_time`
- We compute the execution time
- We set up `prompt_geometry_command_exec_time` if necessary
- Prompt is rendered again (`geometry_prompt_exec_time_render`)

### Case 2: Empty line

- User submits an empty line (ex, `<enter>`)
- No `preexec` hook is triggered as there is no command to run
- `prompt_geometry_command_timestamp` should be undefined as per [line 21](https://gist.github.com/desyncr/4573fa6aea5308aa7813e91ad30e9c7c#file-gistfile1-txt-L21)

Then:

- `precmd` hook is handled by `prompt_geometry_check_command_exec_time`
- We clear up previous values in `prompt_geometry_command_exec_time`
- We compute the execution time (which should end up being `EPOCHSECONDS - EPOCHSECONDS`) as per [this line](https://gist.github.com/desyncr/4573fa6aea5308aa7813e91ad30e9c7c#file-gistfile1-txt-L16)
- We set up `prompt_geometry_command_exec_time` if necessary (not in this case)
- Prompt is rendered again (`geometry_prompt_exec_time_render`)

So, if the case is 2, then this behaviour is correct. There is a room for improvement here as we may explicitly check for `undefined`/`falsy` value to make the logic more obvious.

I'm probably missing a test case as from here it seems to work as expected.

What test case are you testing it against?

cnguy:
I'm so sorry... I've been misunderstanding the plugin's usage this whole time (I was working really late at night most of the times, didn't read properly I think...).  I somehow started to think that the plugin wanted to record the elapsed time since a previous command has finished, and forgot that that the point of it was to display execution times (hence exec_time);.... Yeah, your changes work perfectly then.

Thank you so much for the help over the past week though. I feel like I'm learning a lot even though I've been making a lot of mistakes and overthinking things.

desyncr:
@ChauTNguyen No problem. This happens to me as well as I tend to work on side projects quite late at home.

Well, then there is only matter on copying and pasting [this](https://gist.github.com/desyncr/4573fa6aea5308aa7813e91ad30e9c7c) implementation into this branch.

I'm gonna go ahead and do it. @frmendes Can you review these changes?

desyncr:
@ChauTNguyen Oops. I think since it's a branch on your fork you'll have to merge these changes and pushing them. Otherwise I'll have to close this PR and open up another one. No big deal thou :P

cnguy:
@desyncr Haha, I kinda feel guilty since I've been relying on your work for the past PR's so I'll pass on this one!

desyncr:
@ChauTNguyen No problem! I'm gonna close this up.

Thanks for the time invested in this! Really appreciate it!

PR #76, opened 2016-12-29
Add configuration for git plugin separator

### LINKED ISSUE #52: Complete config for custom separator
Update default plugins to make use of `GEOMETRY_PLUGIN_SEPARATOR` to separate prompt blocks.

Some plugins such as `git` use a hardcoded string as separator (ex. `::`). Review if it's necessary to change it for each plugin or just some of them.

desyncr:
Partially fixes #52 by adding configuration for git plugin.

screen shot 2016-12-29 at 4 36 41 pm

### DIFF
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -77,4 +77,3 @@ PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY=true    # whether to show time if there is n
 GEOMETRY_GIT_NO_COMMITS_MESSAGE="no commits"# message where repository has no commits
 GEOMETRY_GIT_SEPARATOR="::"                 # customize git character separator
 ```
-
--- a/plugins/git/README.md
+++ b/plugins/git/README.md
@@ -48,10 +48,8 @@ GEOMETRY_SYMBOL_GIT_CLEAN="⬢"                 # when reapo has "clean" state
 GEOMETRY_SYMBOL_GIT_REBASE="\uE0A0"           # when in middle of rebase
 GEOMETRY_SYMBOL_GIT_UNPULLED="⇣"              # when there are unpulled changes
 GEOMETRY_SYMBOL_GIT_UNPUSHED="⇡"              # when there are unpushed changes
-GEOMETRY_SYMBOL_GIT_CONFLICTS_SOLVED="◆"      # when all conflicts have been
-solved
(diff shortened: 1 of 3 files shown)


plugins/git/README.md
```diff
@@ -60,7 +58,23 @@ conflicts
...
+PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY=true    # whether to show time if there is no commits
+```
+
+## Misc
+
+```sh
+GEOMETRY_GIT_NO_COMMITS_MESSAGE="no commits"# message where repository has no commits
+GEOMETRY_GIT_SEPARATOR="::"                 # customize git character separator
+```
+
```
frm: Can we remove this extra newline here?

plugins/git/plugin.zsh
```diff
@@ -29,6 +29,7 @@ PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY=${PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY:-true}
 
 # Misc configurations
 GEOMETRY_GIT_NO_COMMITS_MESSAGE=${GEOMETRY_GIT_NO_COMMITS_MESSAGE:-"no commits"}
+GEOMETRY_GIT_SEPARATOR=${GEOMETRY_GIT_SEPARATOR:-"::"}
```
frm, line 32: What do you think about setting a default separator for every plugin and having `GEOMETRY_GIT_SEPARATOR` default to that?

frm (changes requested):

plugins/git/plugin.zsh
desyncr: Technically there is no problem to implement it. My only concern is that it will look a bit cluttered IMO. I would add a config for plugin separators `GEOMETRY_PLUGIN_SEPARATOR` and then leave the option to the user to set both to the same value.

frm: I'm fine with that option. Just the extra newline thing.

desyncr: Ha, it seems `GEOMETRY_PLUGIN_SEPARATOR` already existed. Updated the README.md with a description for it.

plugins/git/README.md
desyncr: Done.

desyncr:
Update `README.md` with minor format fixed and updated a reference to a deprecated configuration.

frm (approved):

PR #77, opened 2016-12-29
Fix exec_time plugin displaying execution time on empty commands

### LINKED ISSUE #72: `exec_time` built-in plugin keeps on being displayed
The `exec_time` built-in plugin shows a strange behavior after being displayed once. It may keep being show whether a command is executed or not.

Steps to reproduce:

- Launch terminal
- Wait at least 5 seconds
- Press Enter key

Expected behavior:

- No `exec_time` plugin is shown

Shown behavior:

- `exec_time` is shown (8s, 9s, etc)

desyncr:
Fixes #72. See #75 for previous discussion.

### DIFF
--- a/plugins/exec_time/plugin.zsh
+++ b/plugins/exec_time/plugin.zsh
@@ -7,27 +7,37 @@ PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME=${PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME:-5
 typeset -g prompt_geometry_command_timestamp
 typeset -g prompt_geometry_command_exec_time
 
-# stores (into prompt_geometry_command_exec_time) the exec time of the last command if set threshold was exceeded
+# Stores (into prompt_geometry_command_exec_time) the exec time of the last command if set threshold was exceeded
 prompt_geometry_check_command_exec_time() {
   integer elapsed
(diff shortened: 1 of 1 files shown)


plugins/exec_time/plugin.zsh
```diff
@@ -7,27 +7,37 @@ PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME=${PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME:-5
...
+  fi
+
+  # Check if elapsed time is above the configured threshold
   (( elapsed = EPOCHSECONDS - ${prompt_geometry_command_timestamp:-$EPOCHSECONDS} ))
   if (( elapsed > $PROMPT_GEOMETRY_COMMAND_MAX_EXEC_TIME )); then
     prompt_geometry_command_exec_time="$(prompt_geometry_seconds_to_human_time $elapsed)"
   fi
+
+  # Clear timestamp after we're done calculating exec_time
+  prompt_geometry_command_timestamp=
```
frm, line 29: I'll approve but do we need to clear the timestamp before and after calculating the time?

frm (approved):

plugins/exec_time/plugin.zsh
desyncr: We only clear `prompt_geometry_command_timestamp` here. And yes it's needed as it's an stale timestamp once used.

PR #83, opened 2017-01-02
Add mercurial (hg) prompt built in plugin

### LINKED ISSUE #82: Mercurial support?
Is it possible to be added? I really like geometry!

desyncr:
Very basic mercurial support:

 - Display current branch
 - Display status symbol (clean, dirty)

Fixes #82

### DIFF
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -4,7 +4,7 @@ typeset -g GEOMETRY_PLUGIN_SEPARATOR=${GEOMETRY_PLUGIN_SEPARATOR:-" "}
 # Define default plugins
 typeset -ga GEOMETRY_PROMPT_PLUGINS
 if [[ $#GEOMETRY_PROMPT_PLUGINS -eq 0 ]]; then
-  GEOMETRY_PROMPT_PLUGINS=(exec_time git)
+  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg)
 fi
 
 # List of active plugins
--- a/plugins/hg/plugin.zsh
+++ b/plugins/hg/plugin.zsh
@@ -22,13 +22,13 @@ geometry_prompt_hg_status() {
 }
 
 geometry_prompt_hg_setup() {
-    return $+commands['hg'];
+  return $+commands['hg'];
 }
 
 geometry_prompt_hg_render() {
-    if [[ -d $PWD/.hg ]]; then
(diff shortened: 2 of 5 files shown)


plugins/hg/plugin.zsh
```diff
@@ -0,0 +1,33 @@
...
+  else
+    echo "$GEOMETRY_HG_CLEAN"
+  fi
+}
+
+geometry_prompt_hg_setup() {}
+
+geometry_prompt_hg_render() {
+  if [[ -d $PWD/.hg ]]; then
+  echo "$(geometry_prompt_hg_branch) ${GEOMETRY_SYMBOL_HG_SEPARATOR} $(geometry_prompt_hg_status)"
```
frm: Indentation

lib/plugin.zsh
```diff
@@ -4,7 +4,7 @@ typeset -g GEOMETRY_PLUGIN_SEPARATOR=${GEOMETRY_PLUGIN_SEPARATOR:-" "}
 # Define default plugins
 typeset -ga GEOMETRY_PROMPT_PLUGINS
 if [[ $#GEOMETRY_PROMPT_PLUGINS -eq 0 ]]; then
-  GEOMETRY_PROMPT_PLUGINS=(exec_time git)
+  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg)
```
frm, line 7: Do we want to have this enabled be default? What if I don't have mercurial installed?

Wouldn't it be possible to at least check if it installed in the setup method?

frm (changes requested):

plugins/hg/plugin.zsh
desyncr: I saw it but it was quite late :)

lib/plugin.zsh
desyncr: Yeah, my idea was to have both working transparently for the user, so if they move between git and mercurial repositories the default prompt will have the same look and feel.

I basically mimicked git plugin that's why I did not put much into mercurial plugin. In this case yeah, it makes sense to have a check.

Edit: grammar

desyncr: ...and yeah it definitely goes into the `setup` phrase.

frm (approved):

desyncr:
@frmendes Should add `time_since_last_commit`?

Edit: My inclination is to leave it as it is as it's on par with other mercurial info prompts, regardless of look and feel between git and mercurial plugins.

desyncr:
I'll go ahead and merge this as it is. Further improvement can be added later on.

PR #86, opened 2017-01-07
Update screenshots

frm:
This took me a little longer than what I wanted. But here are the screenshots.

@desyncr I'd love some feedback particularly regarding [this one](https://github.com/frmendes/geometry/blob/master/screenshots/long_running.png) which I havent updated yet. I'm not really sure what to do without having it end up too empty.


desyncr:
Issuing a couple of long-running commands would be enough I think.

They look great :tada: . Look out for the file size thou.

frm:
File size has to do with retina screens. The previous ones look awful on recent macs.

frm:
@desyncr feedback on that last one?

desyncr:
Looking good :+1:

PR #88, opened 2017-01-07
zsh-async implementation

desyncr:
Branch implementing [zsh-async](https://github.com/mafredri/zsh-async).

Do not merge.

From a quick glance at this, either my zsh-async implementation is doing something wrong or it's noticeably slower than our current implementation.

Re: #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,17 @@ Move the entire `geometry` folder to `$HOME/.oh-my-zsh/custom/themes`, and set `
 
 Add `zplug "frmendes/geometry"` to your `.zshrc`.
 
+### Manual install
+
+Clone this repository as follows:
+
+    git clone https://github.com/frmendes/geometry
+    cd geometry
+    git submodule init && git submodule update
+
+Then add it to your `.zshrc` configuration:
+
+    source /path/to/geometry/geometry.zsh
 
 ## Dependencies
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Clone this repository as follows:
 
     git clone https://github.com/frmendes/geometry
     cd geometry
-    git submodule init && git submodule update
+    git submodule update --init --recursive
 
 Then add it to your `.zshrc` configuration:
 
--- a/lib/async.zsh
+++ b/lib/async.zsh
@@ -1,25 +1,35 @@
 source $GEOMETRY_ROOT/lib/zsh-async/async.zsh
 
--geometry-async-redraw() {
-    # kill any ongoing async_function
-    if [[ "${GEOMETRY_ASYNC_PROC_ID}" != 0 ]]; then
-        kill -- -$(ps -o pgid= $GEOMETRY_ASYNC_PROC_ID | grep -o '[0-9]*') &> /dev/null || :
-    fi
-
-    RPROMPT="${(j/::/)$(prompt_geometry_render_rprompt)}"
-
-    # redisplay
+# Callback handler to properly render RPROMPT with calculated output
+-geometry-async-callback() {
+    local job="$1" code="$2" output="$3" exec_time="$4" stderr="$5"
+    RPROMPT="${(j/::/)output}"
     zle && zle reset-prompt
 }
 
--geometry_async_job() {
-  async_job rprompt_worker -geometry-async-redraw
+# Wrapper to call rprompt renderer, needed to set up workers status
+-geometry-async-prompt() {
+    # In order to work with zsh-async we need to set workers in
+    # the proper directory.
+    cd -q $1 > /dev/null
+    prompt_geometry_render_rprompt
 }
 
-# Hooks precmd to run `-geometry_async_function` and hooks
-# `USR1` signal to comunicate with it
-geometry_async_setup() {
-    async
-    async_start_worker rprompt_worker -u -n
-    add-zsh-hook precmd -geometry_async_job
+# Flushed currently running async jobs and queues a new one
+# See https://github.com/mafredri/zsh-async#async_flush_jobs-worker_name
+-geometry-async-job() {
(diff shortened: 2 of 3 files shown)


lib/async.zsh
```diff
@@ -1,43 +1,29 @@
...
-
-# Renders rprompt and puts result into `${GEOMETRY_ASYNC_TMP_FULL_PATH}$$`
-# then signals `USR1`
--geometry_async_function() {
-    echo "${(j/::/)$(prompt_geometry_render_rprompt)}" > ${GEOMETRY_ASYNC_TMP_FULL_PATH}$$
-    kill -s USR1 $$
-}
+source $GEOMETRY_ROOT/lib/zsh-async/async.zsh
 
 # Launches `-geometry_async_function` keeping track of PID
```
frm: A sidenote: don't forget to rename this in the comment.

frm (changes requested):

frm:
I'm trying to profile this with `zprof`. Can you rebase it off master to better compare both outputs?

frm:
Didn't notice you had rebased, for some reason. I'll do it this weekend.

jedahan:
I have a question - what happens if zsh-async gets sourced by multiple plugins

desyncr:
@jedahan I believe it will just redefine it's own functions and return (see [here](https://github.com/mafredri/zsh-async/blob/master/async.zsh#L473)) -- sorry didn't saw your question before.

@fribmendes Did you gave it a try? Need any help?

frm:
Actually, a small question: are you using anything other than `zprof` to benchmark?

desyncr:
I honestly don't remember. I believe I hacked a couple of `echo's` + time diffs into a file to checkout the difference in performance.

frm:
Ok, so I have spent sometime playing around with `zprof` to no avail. rip time. I then switched to a more manual method.

I measured the time before the [`RPROMPT` variable reset](https://github.com/fribmendes/geometry/blob/master/geometry.zsh#L80) and after the [redraw](https://github.com/fribmendes/geometry/blob/master/lib/async.zsh#L47). Like, I literally only added `echo $(($(gdate +%s%N)/1000000)) >> desyncr.time`. It has a small overhead, yes, but it's not significant for the final results.

In your current solution, in master, it's taking about 5-7ms between each. I am using a small git repo (geometry's repo, actually). When I change to the `zsh-async` solution, it acts weird. There are two redraws. The first takes around 200ms, the second takes an extra 200ms (400ms since the variable reset). It's considerably slower, like you said.

Here's the thing, I'm not sure your implementation is completely correct as I'm not fully familiar with `zsh-async`. However, the current solution is taking 5ms. 5ms to spawn and redraw. Even if we only take into account the first redraw `zsh-async` does (assuming the second is a programming error), the time is still too little to consider adding an external library, IMO. Our solution is self-contained, small and has no submodules for users and developers to handle and google the submodule init command (who even knows it by heart? 😄 ). What do you guys think? @desyncr @jedahan

desyncr:
To measure time here's also: https://unix.stackexchange.com/a/70348

desyncr:
Reworked the implementation. I found out it was being called twice. Fixed the issue and time difference between current `async lib` and `zsh-async` is not noticiable (~0.01).

Implementation has less code and seems clearer. IMHO.

frm:
Are you sure it's ok? Everytime I start a new session, the session simply ends.

desyncr:
Not sure about the session issue, but I found it the last impl. was missing a couple of pieces to properly work. Sorry, not sure how that happened.

frm:
It's working now. On my end, `zsh-async` seems to be just a bit faster (~0.015s). The thing that holds me back is having to rely on other libraries to do the job instead of having everything self-contained. However isn't that much of pain since it's not a big dependency and it is less code to maintain. I'm up for either approach. @desyncr @jedahan what do you guys think?

jedahan:
If I recall correctly, this was motivated due to some bugs with our implementation / maintenance burden of keeping our own. I am ok with relying on zsh-async, and if the maintenance/installation burden is high, switching back to our own implementation.

Two areas of concern:

1. Updating the installation instructions to clone with submodule, which is easy
2. Testing how this would interact with zsh package managers like zpm, zplug, zr, etc.

jedahan (changes requested):
I'd just like to see the readme updated for installing with the new submodule

desyncr:
I have no issue with either of those implementations. I guess I'd go with zsh-async as to rely on an already proven solution.

Will update `README.md` will manual install instructions.

desyncr:
Done. Ready for final review.

desyncr:
With this merged and background job plugin we can call it 1.1.0?

frm (approved):

frm:
I think so. Can we have a chat on gitter on roadmapping and "marketing" (as we mentioned in #146)?

desyncr:
@fribmendes I'm up for it. Will be on and off anyways.

README.md
```diff
@@ -36,6 +36,17 @@ Move the entire `geometry` folder to `$HOME/.oh-my-zsh/custom/themes`, and set `
 
 Add `zplug "frmendes/geometry"` to your `.zshrc`.
 
+### Manual install
+
+Clone this repository as follows:
+
+    git clone https://github.com/frmendes/geometry
+    cd geometry
+    git submodule init && git submodule update
```
jedahan: how do you feel about `git submodule update --init --recursive`

jedahan (approved):
One small request but looks ready to merge with or without a change

README.md
desyncr: I haven't used it for `zsh-async` as it doesn't have submodules. Future proof maybe?

desyncr: ``git submodule update --init`` seems fine to be. Gonna update later today.

PR #89, opened 2017-01-09
Update oh-my-zsh instructions

cnguy:
Hello again!

In the case that users already have themes within the custom folder, I moved the folder (without the .git stuffs) to not overwrite anything, which is also why I set `ZSH_THEME` to `geometry/geometry`. Simply moving geometry.zsh and renaming it does not work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,7 @@ Just add `antigen bundle frmendes/geometry` to your `.zshrc`.
 
 ### oh-my-zsh
 
-Move `geometry.zsh` to `$HOME/.oh-my-zsh/custom/themes/geometry.zsh-theme` and
-set `ZSH_THEME="geometry"` in your `.zshrc`.
+Move the entire `geometry` folder to `$HOME/.oh-my-zsh/custom/themes`. Rename `geometry.zsh` to `geometry.zsh-theme` and set `ZSH_THEME="geometry/geometry"` in your `.zshrc`.
 
 ### zplug
 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Just add `antigen bundle frmendes/geometry` to your `.zshrc`.
 
 ### oh-my-zsh
 
-Move the entire `geometry` folder to `$HOME/.oh-my-zsh/custom/themes`. Rename `geometry.zsh` to `geometry.zsh-theme` and set `ZSH_THEME="geometry/geometry"` in your `.zshrc`.
+Move the entire `geometry` folder to `$HOME/.oh-my-zsh/custom/themes`, and set `ZSH_THEME="geometry/geometry"` in your `.zshrc`.
 
 ### zplug
 
--- a/geometry.zsh-theme
+++ b/geometry.zsh-theme
@@ -1,0 +1,1 @@
+geometry.zsh


README.md
```diff
@@ -26,8 +26,7 @@ Just add `antigen bundle frmendes/geometry` to your `.zshrc`.
 
 ### oh-my-zsh
 
-Move `geometry.zsh` to `$HOME/.oh-my-zsh/custom/themes/geometry.zsh-theme` and
-set `ZSH_THEME="geometry"` in your `.zshrc`.
+Move the entire `geometry` folder to `$HOME/.oh-my-zsh/custom/themes`. Rename `geometry.zsh` to `geometry.zsh-theme` and set `ZSH_THEME="geometry/geometry"` in your `.zshrc`.
```
frm: Does it really have to rename it? `oh-my-zsh` assumes that file extension?

cnguy: `/Users/<name>/.oh-my-zsh/oh-my-zsh.sh:source:110: no such file or directory: /Users/<name>/.oh-my-zsh/themes/geometry.zsh-theme`

I tried it both on my mac & linux comps and both times I had to use the `zsh-theme` extension.

small kinda irrelevant note: in the above error output I have geometry's contents in the top-level and not in a geometry folder so it didn't have the error as `themes/geometry/geometry.zsh-theme`.

cnguy: https://github.com/robbyrussell/oh-my-zsh/wiki/Customization

> Overriding and adding themes
> 
> Adding and customizing your own themes pretty much works the same as with plugins.
> 
> Themes are located in a themes folder and must end with .zsh-theme. The basename of the file is the name of the theme.

desyncr (approved):

README.md
frm: I'm really more of a fan of minimal configuration and having the users do little to no effort to install it. How about we create a symlink? `geometry.zsh-theme -> geometry.zsh`? It's the way `geometry.plugin.zsh` is.

What's your opinion?

cnguy: I didn't know what a symlink was until you mentioned it to me haha. I just committed a change. Are my original instructions still okay though (besides the fact that I can now delete the second statement)?

frm: @ChauTNguyen with the symlink you can now change that rename step, right?

cnguy: I can just delete it actually!

cnguy: Woops, delete the first half.* my bad

frm: Yes, you're right! Thank you very much! When you update the README I'll merge everything.

frm (approved):

desyncr:
Just a note: with a `.zsh-theme` now antigen instructions can be updated to:

`antigen theme frmendes/geometry`

frm:
@ChauTNguyen want to make a new PR?

cnguy:
I'm all for it! Is there a way to combine it with the last one to make it cleaner since that one will only have one change (if you prefer it that way). Or should I make another one?

frm:
The other one is already merged. There's no way to combine it without force pushing to master (which rewrites history). You can just make a new one.

cnguy:
I understand. PR: https://github.com/frmendes/geometry/pull/90

Issue #57, opened 2016-12-10
Add Node/yarn version plugin


desyncr:
Create built-in plugin (in the lines of `virtualenv` and `docker_machine`) for `node` and `npm` version prompt.

Prompt should look something like:

    ⬡ v6.9.2 (3.10.9)

`npm`s start up time is quite noticeable so version information should be cached in the setup phrase, in fact, this should be done for `node` and `npm`.

This is a great opportunity for anyone interested in contributing.

castellanprime:
Hi. First time contributor here. Machine is Debian 8. Could I take this issue?

desyncr:
Hi @castellanprime Sure thing. Go ahead and ask any question you have :+1:

castellanprime:
So is it in the same format as docker-machine and virtualenv?

desyncr:
It should be done in the form of a [plugin](https://github.com/frmendes/geometry/tree/master/plugins#plugins), the format should be:

    ⬡ v6.9.2 (3.10.9)

We're open to suggestions on what to show. I'm not a node developer so I initially proposed the format above, but if there is other more useful information to show, go ahead.

My suggestion is to first start with the plugin structure (setup/render methods) and little more than that. Once you get this done then add in the `render` method a basic `echo $(node -v)` and we'll move from there :)

castellanprime:
Okay. Doing that now.

castellanprime:
What is the version number inside the brackets?

desyncr:
Should be `npm --version` IIRC.

frm:
It's the npm version but I'd recommend supporting [yarn](https://yarnpkg.com/) when a `yarn.lock` is present.

castellanprime:
Should we also support nodeenv and nvm as well? I installed node using nvm.

desyncr:
I guess we can add a prefix if the active node version is from nvm.

With nvm installed:

    ⬡ nvm: v6.9.2 (3.10.9)

vs without nvm:

    ⬡ v6.9.2 (3.10.9)

No need for `nodeenv` for now.

Don't try to add every single tool available, with a basic plugin running that's enough, we can further improve it later on :)

castellanprime:
Okay. I can print the versions but I have to add some code to detect whether the node is from yarn or from nvm.

desyncr:
I'd suggest to push your working code and making a PR, that way we can start the code review and testing it out.

castellanprime:
Okay. Doing that now

[...]

PR #91, opened 2017-01-12
Added Node/npm version plugin

castellanprime:
This is version 1. it is still missing the enhancements talked out in the last commit. #57

### DIFF
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -4,7 +4,7 @@ typeset -g GEOMETRY_PLUGIN_SEPARATOR=${GEOMETRY_PLUGIN_SEPARATOR:-" "}
 # Define default plugins
 typeset -ga GEOMETRY_PROMPT_PLUGINS
 if [[ $#GEOMETRY_PROMPT_PLUGINS -eq 0 ]]; then
-  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg)
+  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg node)
 fi
 
 # List of active plugins
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -4,7 +4,7 @@ typeset -g GEOMETRY_PLUGIN_SEPARATOR=${GEOMETRY_PLUGIN_SEPARATOR:-" "}
 # Define default plugins
 typeset -ga GEOMETRY_PROMPT_PLUGINS
 if [[ $#GEOMETRY_PROMPT_PLUGINS -eq 0 ]]; then
-  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg node)
+  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg)
 fi
 
 # List of active plugins
--- a/plugins/node/README.md
+++ b/plugins/node/README.md
@@ -4,6 +4,13 @@ Node plugin. Displays node and npm version.
 
 ## Configuration
 
+### Colors
+
+```sh
+GEOMETRY_COLOR_NODE_NPM_VERISON="black"
+```
+
+
 ### Symbols
 
 ```sh
--- a/plugins/node/README.md
+++ b/plugins/node/README.md
@@ -7,7 +7,7 @@ Node plugin. Displays node and npm version.
 ### Colors
 
 ```sh
-GEOMETRY_COLOR_NODE_NPM_VERISON="black"
+GEOMETRY_COLOR_NODE_NPM_VERSION="black"
 ```
 
 
--- a/plugins/node/README.md
+++ b/plugins/node/README.md
@@ -1,6 +1,6 @@
 # Node
 
-Node plugin. Displays node and npm version.
+Node plugin. Displays node and npm version only.
 
 ## Configuration
 
--- a/plugins/node/README.md
+++ b/plugins/node/README.md
@@ -1,0 +1,11 @@
+# Node
+
+Node plugin. Displays node and npm version.
+
+## Configuration
+
+### Symbols
+
+```sh
+GEOMETRY_SYMBOL_NODE_NPM_VERSION="⬡"
+```
--- a/plugins/node/plugin.zsh
+++ b/plugins/node/plugin.zsh
@@ -1,18 +1,24 @@
 # Color definitions
-GEOMETRY_COLOR_NODE_NPM_VERISON=${GEOMETRY_COLOR_NODE_NPM_VERISON:-black}
+GEOMETRY_COLOR_NODE_NPM_VERSION=${GEOMETRY_COLOR_NODE_NPM_VERSION:-black}
 
 # Symbol definitions
-GEOMETRY_SYMBOL_NODE_NPM_VERISON=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
-GEOMETRY_NODE_NPM_VERISON=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERISON GEOMETRY_SYMBOL_NODE_NPM_VERISON) 
+GEOMETRY_SYMBOL_NODE_NPM_VERSION=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERSION $GEOMETRY_SYMBOL_NODE_NPM_VERSION) 
 
-geometry_prompt_node_setup() 
+local node='node'
+local npm='npm'
+
+geometry_prompt_node_setup(){} 
 
 geometry_prompt_node_render(){
-	local node_path="$echo "$(which node)" -v"
-	local npm_path="$echo "$(which npm)" --version"
-	GEOMETRY_NODE_VERSION=`(eval ${node_path})`
-	GEOMETRY_NPM_VERSION=`(eval ${npm_path})`
-	echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
+	local n=$+commands[$node]
+	if (( $n == 0 )); then
+		echo ""
+	else
+		GEOMETRY_NODE_VERSION="$($node -v)"
+		GEOMETRY_NPM_VERSION="$($npm --version)" 
+		echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
+	fi
 }
 
 geometry_plugin_register node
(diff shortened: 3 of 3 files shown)


desyncr:
Looks great to me :+1:. Just gonna add a few comments.

lib/plugin.zsh
```diff
@@ -4,7 +4,7 @@ typeset -g GEOMETRY_PLUGIN_SEPARATOR=${GEOMETRY_PLUGIN_SEPARATOR:-" "}
 # Define default plugins
 typeset -ga GEOMETRY_PROMPT_PLUGINS
 if [[ $#GEOMETRY_PROMPT_PLUGINS -eq 0 ]]; then
-  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg)
+  GEOMETRY_PROMPT_PLUGINS=(exec_time git hg node)
```
desyncr: Not sure about adding it as a default built-in plugin. @frmendes What do you think?

plugins/node/README.md
```diff
@@ -0,0 +1,18 @@
+# Node
+
+Node plugin. Displays node and npm version.
+
+## Configuration
+
+### Colors
+
+```sh
+GEOMETRY_COLOR_NODE_NPM_VERISON="black"
```
desyncr: Typo: `VERISON`  -> `VERSION`.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,18 @@
+# Color definitions
+GEOMETRY_COLOR_NODE_NPM_VERISON=${GEOMETRY_COLOR_NODE_NPM_VERISON:-black}
```
desyncr: Typo same as above.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,18 @@
...
+
+# Symbol definitions
+GEOMETRY_SYMBOL_NODE_NPM_VERISON=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERISON=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERISON GEOMETRY_SYMBOL_NODE_NPM_VERISON) 
+
+geometry_prompt_node_setup()
```
desyncr: Should be `geometry_prompt_node_setup() {}` in order to define the function correctly.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,18 @@
...
+# Symbol definitions
+GEOMETRY_SYMBOL_NODE_NPM_VERISON=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERISON=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERISON GEOMETRY_SYMBOL_NODE_NPM_VERISON) 
+
+geometry_prompt_node_setup() 
+
+geometry_prompt_node_render(){
+	local node_path="$echo "$(which node)" -v"
+	local npm_path="$echo "$(which npm)" --version"
+	GEOMETRY_NODE_VERSION=`(eval ${node_path})`
```
desyncr: There are a couple of issues here but my suggestion is to keep it simple:

1- Check node/npm are available
2- Display version info

For 1) zsh has a neat way to do it:

`$+commands[node]` (or for `npm`: `$+commands[npm]`).


You should check if there is no node/npm command available and return an empty string if none is available:

    if (( ! $+commands[node] )); then
        return "";
    fi


For 2) you can simply do:

    GEOMETRY_NODE_VERSION="$(node -v)"

desyncr (changes requested):
There are a couple of issues but great work so far! :)

plugins/node/README.md
castellanprime: Okay.

plugins/node/plugin.zsh
castellanprime: Will change

lib/plugin.zsh
castellanprime: I will remove it from the default built-in plugin.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,18 @@
+# Color definitions
+GEOMETRY_COLOR_NODE_NPM_VERISON=${GEOMETRY_COLOR_NODE_NPM_VERISON:-black}
+
+# Symbol definitions
+GEOMETRY_SYMBOL_NODE_NPM_VERISON=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERISON=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERISON GEOMETRY_SYMBOL_NODE_NPM_VERISON)
```
desyncr: Here you are missing `$` on `GEOMETRY_SYMBOL_NODE_NPM_VERISON`. That's the reason the symbol is not displayed.

desyncr (changes requested):
One more comment :)

plugins/node/plugin.zsh
castellanprime: Thanks. Will change now.

plugins/node/plugin.zsh
castellanprime: From your explanation, the statement "$+commands[node]" returns an alias "node". Is this right? If so, how to add npm to this array?

desyncr: `$+commands[node]` will return `1` if it finds `node` in the `commands` array. Example:

With `node` command available in `$PATH`:

    > echo $+commands[node]
    1

Without `node` available:

    > echo $+commands[node]
    0

If node is found in `$+commands` we should be able to do `node -v`, otherwise I'd return immediately without displaying anything.

lib/plugin.zsh
frm: I'd rather not have it as a default plugin, to keep the "minimal default" philosophy

castellanprime:
I am sorry that I have not committed anything towards this in the last few days. I have been very busy. I will work on the code from tonight.

desyncr:
@castellanprime awesome. Don't worry. Keep up the good work!

castellanprime:
I have fixed the errors in the pull request.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,24 @@
+# Color definitions
+GEOMETRY_COLOR_NODE_NPM_VERSION=${GEOMETRY_COLOR_NODE_NPM_VERSION:-black}
+
+# Symbol definitions
+GEOMETRY_SYMBOL_NODE_NPM_VERSION=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERSION $GEOMETRY_SYMBOL_NODE_NPM_VERSION) 
+
+local node='node'
```
desyncr: No need to put it into a variable.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,24 @@
...
+local npm='npm'
+
+geometry_prompt_node_setup(){} 
+
+geometry_prompt_node_render(){
+	local n=$+commands[$node]
```
desyncr: Same here. No need to use a variable. Better  use:

    if (( $+commands ['node'] == 0 ))

desyncr (changes requested):
Great! Just a few changes.

Also use 4 spaces for indentation.

castellanprime:
Added your changes

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,21 @@
...
+
+# Symbol definitions
+GEOMETRY_SYMBOL_NODE_NPM_VERSION=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERSION $GEOMETRY_SYMBOL_NODE_NPM_VERSION) 
+
+
+geometry_prompt_node_setup(){} 
+
+geometry_prompt_node_render(){
+    if (( $+commands ['node'] == 0 )); then
```
desyncr: I think I gave you a bad tip. This line should read as follow:

    if (( $+commands[node] == 0 )); then

So, instead of `'node'` (quoted), it should be `node`.

desyncr (changes requested):
I gave you a wrong example. My bad.

This small fix up and we'll be done and ready! Cheers!

plugins/node/plugin.zsh
castellanprime: Fixed that. Just pushed. Thanks for your help so far.

castellanprime: I am working on the detecting the yarn.lock file and .nvmrc file but I am running into problems. This is the line I want to add:
`if [[-e yarn.lock]] || [[ -f .nmvrc && -r .nvrmc && -n "$(head -1 .nvmrc)]]; then"`.
I want add this above the 
`if (( $+commands[node] == 0)); then`

I tested it using a simple .nvmrc file in a directory jsproj using this steps 
screenshot from 2017-01-23 22 58 29

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,21 @@
...
+
+# Symbol definitions
+GEOMETRY_SYMBOL_NODE_NPM_VERSION=${GEOMETRY_SYMBOL_NODE_NPM_VERSION:-"⬡"}
+GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VERSION $GEOMETRY_SYMBOL_NODE_NPM_VERSION) 
+
+
+geometry_prompt_node_setup(){} 
+
+geometry_prompt_node_render(){	
+    if (( $+commands[node] == 0 )); then
```
desyncr: Remember that `$+commands[node]` will return `0` if the command does not exists, `1` otherwise. So, this conditional is backwards.

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,21 @@
...
+        GEOMETRY_NODE_VERSION="$(node -v)"
+        GEOMETRY_NPM_VERSION="$(npm --version)" 
+        echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
+    else
+        echo " "
```
desyncr: In case there is no node available do not generate an empty space, ie: `echo ""`.

desyncr (changes requested):
Getting closer. Review the conditional and an empty space and we'll be done!

castellanprime:
Added your changes. I wonder if you looked at my earlier suggestion about the .nmvrc/yarn.lock files?  Is that what you wanted? Thanks.

desyncr (approved):
Very happy with your work! Gonna merge this now!

desyncr:
@castellanprime Regarding `yarn.lock` I think it's a great addition to put us ahead of other prompts. If you are up for it we can work on this on another PR. :)

Issue #58, opened 2016-12-10
Add Ruby/rvm version plugin 


desyncr:
Similar to #57 create a plugin for ruby/rvm version information prompt.

Use `rvm`/`rbenv` if available. Display in the following format:

     ◆ 2.3.3 (1.27.0)

Cache version information in the setup phrase.

This is a great opportunity for anyone interested in contributing.

frm:
I have a set up ready for this, might merge if no one else wants to tackle it.

reuben453:
Hi, I'd like to give this a shot. Can I take this issue?

reuben453:
It will probably be useful if the current rvm gemset is also displayed. What do you think?

desyncr:
@reuben453 I'd have it under a configuration variable, something like:

    GEOMETRY_RUBY_RVM_SHOW_GEMSET=[true|false]

Good work so far in #93

reuben453:
Should this be closed now?

frm:
@reuben453 you're right, closing.

PR #93, opened 2017-01-14
Add ruby plugin to display ruby and rvm version information

reuben453:


reuben453:
Fixes #58

reuben453:
Haven't added caching of the version info in the setup as yet.

desyncr:
Gonna have a look at this right now :+1:

desyncr (approved):
Looks great! It's ready to merge IMO. I'll leave @frmendes  to final review.

reuben453:
Added caching and display of the current rvm gemset name.
The `GEOMETRY_RUBY_RVM_SHOW_GEMSET` variable is used to control whether the gemset name is displayed or not. It defaults to true.

So it now displays like this: ◆ 2.3.0p0 (1.27.0 default)
where default is the current gemset.

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,65 @@
...
+
+  if (( $+commands[rvm] )); then
+      get_full_rvm_version
+      [[ $GEOMETRY_RVM_VERSION_FULL =~ 'rvm ([0-9a-zA-Z.]+)'  ]]
+      GEOMETRY_RVM_VERSION=$match[1]
+      result=$result" ($GEOMETRY_RVM_VERSION"
+
+      # Add current gemset name
+      if $GEOMETRY_RUBY_RVM_SHOW_GEMSET; then
+          cur_dir=$(pwd)
```
desyncr: I'd move all this branched code inside a function.

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,65 @@
...
+
+  GEOMETRY_RVM_VERSION_FULL="$(rvm -v)"
+}
+
+geometry_prompt_ruby_render() {
+
+  if (( ! $+commands[ruby] )); then
+      return "";
+  fi
+  get_full_ruby_version
```
desyncr: Why it's necessary to call this function on each render cycle. Can't it be done once in the setup phase?

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,65 @@
...
+}
+
+get_full_rvm_version() {
+  if (( $+commands[rvm] )) && [[ -z $GEOMETRY_RVM_VERSION_FULL ]]; then
+      GEOMETRY_RVM_VERSION_FULL="$(rvm -v)"
+  fi
+}
+
+geometry_prompt_ruby_setup() {
+  GEOMETRY_RUBY_VERSION_FULL="$(ruby -v)"
```
desyncr: Can't we use get_full_ruby_version function for this? To avoid code duplication.

desyncr (changes requested):
Few comments. Overall great work! Thanks a lot!

reuben453:
@desyncr I've changed the code based on your comments. 

I'm not very clear on when the setup and render functions are called. I assume the setup is run once/rarely, while the render is run every time the prompt needs to be rendered.

plugins/ruby/plugin.zsh
```diff
@@ -5,25 +5,39 @@ GEOMETRY_COLOR_RUBY_RVM_VERSION=${GEOMETRY_COLOR_PROMPT:-white}
...
+
+  GEOMETRY_RVM_VERSION_FULL="$(rvm -v)"
```
frm: Can we use `get_full_rvm_version`?

frm (changes requested):

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+get_full_rvm_version() {
+  if (( $+commands[rvm] )) && [[ -z $GEOMETRY_RVM_VERSION_FULL ]]; then
+      GEOMETRY_RVM_VERSION_FULL="$(rvm -v)"
+  fi
+}
+
+geometry_prompt_ruby_setup() {
+  get_full_ruby_version
```
frm: The `setup` function can be empty, right?

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,72 @@
...
+      else
+          return $match[1]
+      fi
+  else
+    return ""
+  fi
+}
+
+geometry_prompt_ruby_render() {
+
```
frm: Unnecessary newline here.

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,72 @@
...
+  if (( ! $+commands[ruby] )); then
+      return "";
+  fi
```
frm, line 52: Can you add a newline here?

frm (changes requested):

frm:
@reuben453 Thanks! I commented on some code changes to make it consistent.

Another thing would be the function naming. You can see [here](https://github.com/frmendes/geometry/blob/master/plugins/git/plugin.zsh) that were are using `prompt_geometry` as prefix for functions used by plugins. Can you update that please?

As for your question, the `setup` function gets run once when the theme is loaded. Plugins can use it to do checks and configure anything they require. Imagine you want to determine what file pattern searcher the user has. It can be `grep`, `ag` or `rg`. You can just check which one is installed and save that into a `FILE_SEARCHER` variable to use later in the plugin. `render` gets run everytime the prompt is going to be rendered.

plugins/ruby/plugin.zsh
reuben453: I thought it would be better to run `ruby -v` and `rvm -v` in the setup.

plugins/ruby/plugin.zsh
reuben453: Removed

plugins/ruby/plugin.zsh
reuben453: Added

reuben453:
Thanks for the explanation @frmendes!

I've changed the prefix for the functions.

plugins/ruby/plugin.zsh
frm: Right, but you run it again on `render`. And you have to do so because there might be several ruby versions in different projects, right? So running on `setup` isn't going to do much. Or am I understanding it wrong?

reuben453: Hmm, you're right, I've updated the code.

reuben453:
The git plugin uses `geometry_prompt` as the prefix for the setup and render functions, while the other functions use `prompt_geometry` as the prefix. Should I do the same?

frm:
@reuben453 that would be ideal, yes. Don't forget that the `setup` step is not required and can simply be an empty function.

reuben453:
Changed the prefixes and the setup function.

frm (approved):
Looks good to me. @desyncr any take on this?

plugins/ruby/plugin.zsh
```diff
@@ -0,0 +1,68 @@
...
+}
+
+geometry_prompt_ruby_render() {
+  if (( ! $+commands[ruby] )); then
+      return "";
+  fi
+
+  prompt_geometry_get_full_ruby_version
+  [[ $GEOMETRY_RUBY_VERSION_FULL =~ 'ruby ([0-9a-zA-Z.]+)' ]]
+  GEOMETRY_RUBY_VERSION=$match[1]
```
desyncr: These two lines could be moved into a function such as 'prompt_geometry_ruby_version'. Same for rvm.

desyncr (approved):
Excellent work. A small non functional nitpick. I'd merge this as it's, though, as it's a great addition.

reuben453:
@desyncr, moved those lines into separate functions.

desyncr (approved):

frm:
We both approved, merging. Thanks again!

PR #96, opened 2017-01-26
Clarify plugins used in screenshots

### LINKED ISSUE #92: Clarify plugins used in screenshots
Some users may assume that syntax highlighting will be enabled with the theme (command colouring)  - at least I got asked once.

So it may be useful to add a note explaining which plugins are used.

AFAIK there is only `zsh-users/zsh-syntax-highlighting` plugin installed in the screenshots.

Edit: Other thing to add would be the terminal and `rupa/z` plugin.

frm:
Fixes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,12 @@ As for PRs, we are currently looking to improve performance.
 
 If you want to help with plugins, we are looking to add Ruby and Node version displays.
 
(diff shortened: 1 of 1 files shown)


desyncr:
Done! :D @frmendes

PR #98, opened 2017-01-27
Create rustup plugin

jedahan:
It adds a little gear icon to the right, that changes color for the three rust toolchains, like so:

screen shot 2017-01-27 at 1 49 28 am

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -12,6 +12,7 @@ Available plugins:
 * [Hg](/plugins/hg)
 * [Node](/plugins/node)
 * [Ruby](/plugins/ruby)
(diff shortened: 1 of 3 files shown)


frm (approved):
I really, really like this. Merging. Thanks!

frm:
Btw, you contributed to geometry before and after the refactor. Do you have any feedback you would like to share?

jedahan:
So I think you are doing a very good job balancing simplicity with extendability. Often my favorite projects grow to unmanageable complexity (like zplug), but having strong opinions on what a plugin is (asynchronous suggestion in RPROMPT only) has been good to see. I will think more about it soon.

PR #99, opened 2017-01-28
Update plugins/rustup with two minor fixes

jedahan:
The first is making the naming consistent, the second is to only hide the separator if we are in a non-git-managed directory.


desyncr (approved):
Looking good. Merging.

PR #103, opened 2017-02-04
They looked like multiple steps.

Indyandie:
Just to make it a little clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,6 @@ geometry starts small, with good defaults, and allows you to customize it at you
 
 *K, I'm sold. Beam me up, Scotty.*
 
-<<<<<<< HEAD
 ### Install using antigen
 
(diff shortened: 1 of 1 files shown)


desyncr:
What about:

    Option 1: Using antigen
    Option 2: Using oh-my-zsh
    Option 3: Using zplug

Could we give it a try?

Indyandie:
how about 

`### Install using <method>`

desyncr:
Yeah, much better.

desyncr:
@Indyandie Are you up to implement the format you described [above](https://github.com/frmendes/geometry/pull/103#issuecomment-277501033)? Let me know :)

Indyandie:
@desyncr okay we should be all good now 👍🏾👍🏾

desyncr (approved):
Looking good!

PR #104, opened 2017-02-07
widen prompt_geometry_hash_color range

### LINKED ISSUE #87: Random hostname colouring range
As of now it's not really that much random.

There are a lot of recurrent colors (purple, blue, light blue). Might have to do with my own setup which currently is [Base16 Tomorrow Night](https://github.com/chriskempson/base16).

However there seem to be some colors that are missing out, mainly to try and avoid conflicts with the red symbol used when a command has a non-zero exit.

jedahan:
This should fix #87 - though there will be a lot more color 'collisions' depending on your eyesight, theme, etc. 

For example, these are colors 1-256 on tomorrow night eighties.

screen shot 2017-02-06 at 9 07 45 pm

### DIFF
--- a/lib/color.zsh
+++ b/lib/color.zsh
@@ -4,10 +4,10 @@ prompt_geometry_colorize() {
 }
 
 prompt_geometry_hash_color() {
-  colors=(2 3 4 6 9 12 14)
+  colors=(`seq 1 9`)
 
   if (($(echotc Co) == 256)); then
-    colors+=(99 155 47 26)
+    colors+=(`seq 17 230`)
   fi
 
   local sum=0


lib/color.zsh
```diff
@@ -4,10 +4,10 @@ prompt_geometry_colorize() {
 }
 
 prompt_geometry_hash_color() {
-  colors=(2 3 4 6 9 12 14)
+  colors=(`seq 1 9`)
```
desyncr, line 7: I believe we tend yo use $()  rather than backticks.

desyncr (changes requested):

frm:
From what I'm seeing, this covers the whole colorspace, right? The apparent lack of yellows comes down to the user theme and not issues with hashing.

jedahan:
This does not cover the whole colorspace. It skips 10-16 and 232-256 because those seem to be slightly different shades of gray which I feel are harder to distinguish. This may totally be my colorscheme specific, and would love to see a few others print out colors 1-256 so we can make a better decision

lib/color.zsh
frm: It's defined inside (). Doing `colors=($(seq 1 9))` can seem a bit confusing. I'm fine with both.

frm:
This is the `seq` output for geometry with base16 tomorrow night. 

screen shot 2017-02-07 at 19 18 54

I must've screwed up something in the output but it seems to be fair in terms of colors.

desyncr (approved):

PR #106, opened 2017-02-07
Enable user to extend/customize lprompt

### LINKED ISSUE #51: Left prompt customization
- Modify prompt to make use of plugins
- Be able to configure plugins for rprompt as well as prompt.
- Evaluate which customization can be added

desyncr:
Partially fixes #51. This change enable users to overwrite/extend left prompt by creating the function `prompt_geometry_render_lprompt`.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -57,6 +57,10 @@ prompt_geometry_render_rprompt() {
     geometry_plugin_render
 }
 
+prompt_geometry_render_lprompt() {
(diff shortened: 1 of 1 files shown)


frm (approved):
LGTM but I do question what's next for this. As I mentioned in Gitter, I'd rather not deconstruction the left prompt plugin too much as that's a very personal choice for users. It would get too complex and hard to maintain quickly. This is the most straightforward option and it's good, of course. But I was wondering if you have anything besides that in mind.

desyncr:
@frmendes I was messing with this yesterday. My same thought, deconstructing left prompt will not be too useful. I could not finish transforming left prompt into a plugin format.

This will enable users to use current built-in plugins in the left or right prompt, add custom functions and such.

I'll give it a try and we'll see the cost-benefits of it.

PR #107, opened 2017-02-07
[WIP] Compilation step

### LINKED ISSUE #50: Compile step of default plugins into geometry
The idea is to compile geometry.zsh with `lib/*` and `plugins/*` bundled into it in order to have a single `.zsh` to source.

desyncr:
Kind of working compilation script.

### Issues:

  - [x] Built-in plugins all activated
  - [x] No version information in `bin/geometry.zsh`

This script seems too hacky. It's an attempt to only relay on compilation (rather than run-time checks, ie env variables).

Should fix #50


frm:
I'm not sure we should merge on the current form. As we spoke on #50, there should be sourcing depending on a `GEOMETRY_ENV` variable, shouldn't it?

Also, built-in plugins all activated?

desyncr:
Yep It's a work in progress. It should not be merged. It's just to show current status.

desyncr:
```
 ▲ ~/Projects/geometry bin/compile
Removing target: /home/user/Projects/geometry/build/geometry.zsh
Compiling into a single file...

--> Disable self-register
--> Register default plugin
--> Disable sourcing lib

Built: /home/user/Projects/geometry/build/geometry.zsh
Done!
```

desyncr:
Workflow would be as follows:

  - Build `build/geometry.zsh` when creating a new release
  - On the release artifacts offer `build/geometry.zsh`

Plugin managers may use `geometry.zsh` or `build/geometry.zsh`, it works the same way, ie:

    antigen theme frmendes/geometry
    # or antigen theme frmendes/geometry build/geometry

For development the user should use `geometry.zsh` from the repository (not the compiled version) -- no need for environment variable.

jedahan:
Is there a significant startup difference between 1 file and multiple files? In general the idea seems fine to me, but I am not sure what the goals are, and it does add some complexity to the project.

desyncr:
There is no major time difference between one file or multiple. The idea is simplify manual install.

desyncr:
Maybe there is no need for this as of now. Should I close this? @fribmendes @jedahan

jedahan:
I would close this. I see no need for a compilation step yet.

PR #108, opened 2017-02-08
Added npm/yarn check for node plugin

castellanprime:
Some bugs in the implementation. Prompt does not print. #57


desyncr:
I think you have to [`rebase`](http://gitready.com/intermediate/2009/01/31/intro-to-rebase.html). Let me know you need some help!

plugins/node/plugin.zsh
```diff
@@ -0,0 +1,39 @@
...
+geometry_prompt_node_render(){
+    local -a files
+    files=($(ls -a ${PWD} | egrep -e ".nvmrc|yarn.lock"))
+
+    if [[ -n $files[1] ]]; then
+        case "$files[1]" in 
+            ".nvmrc")
+                local NODE_VERSION=$(nvm version "$(cat $files[1])")
+                if [[ "$NODE_VERSION" != "N/A" ]]; then
+                    nvm use 2> /dev/null
```
frm: I'm not sure how I feel about this. If I understand it correctly, and please correct me if I'm wrong, but it makes changes to the environment instead of being idempotent and just reading the version. It assumes the user wants to use the version specified on `.nvmrc` which is not something we should feel comfortable about enforcing.

Does anyone else have takes on this?

frm (changes requested):

plugins/node/plugin.zsh
desyncr: Just echoing `$NODE_VERSION` is enough.

desyncr (changes requested):

castellanprime:
@desyncr I ran this command(git pull --rebase https://github.com/frmendes/geometry.git master). Is that the proper command?

castellanprime:
@desyncr I was overthinking this. The code now works!. It only prints the version when an .nvmrc file or yarn file is found. Thanks @desyncr

desyncr:
Thank you @castellanprime for your amazing work!

plugins/node/plugin.zsh
```diff
@@ -8,11 +8,17 @@ GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VE
...
 geometry_prompt_node_render() {
-    if (( $+commands[node] )); then
-        GEOMETRY_NODE_VERSION="$(node -v 2> /dev/null)"
-        GEOMETRY_NPM_VERSION="$(npm --version 2> /dev/null)" 
-        echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
+    local -a files
+    files=($(ls -a ${PWD} | egrep -e ".nvmrc|yarn.lock"))
+
+    if [[ ${#files[@]} -gt 0 ]]; then
+        if (( $+commands[node] )); then
```
desyncr: I'd move this check above all inside this function. Returning immediately if there is no `node` command available. This way we have a clearer intent:

    # Return immediately if there is no `node` command available.
    if [[ ! $+commands[node] ]]; then
        return
    fi

plugins/node/plugin.zsh
```diff
@@ -8,11 +8,17 @@ GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VE
 geometry_prompt_node_setup() {}
 
 geometry_prompt_node_render() {
-    if (( $+commands[node] )); then
-        GEOMETRY_NODE_VERSION="$(node -v 2> /dev/null)"
-        GEOMETRY_NPM_VERSION="$(npm --version 2> /dev/null)" 
-        echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
+    local -a files
+    files=($(ls -a ${PWD} | egrep -e ".nvmrc|yarn.lock"))
```
desyncr: I would rather use `[[ -f $FILE ]]` instead of an array. It makes it easier to read and we are checking against files we already know the names:

    if [[ -f "$PWD/.nvmrc" || -f "$PWD/yarn.lock" ]]; then # if .nvmrc or yarn.lock exists...

plugins/node/plugin.zsh
```diff
@@ -8,11 +8,17 @@ GEOMETRY_NODE_NPM_VERSION=$(prompt_geometry_colorize $GEOMETRY_COLOR_NODE_NPM_VE
...
+
+    if [[ ${#files[@]} -gt 0 ]]; then
+        if (( $+commands[node] )); then
+            GEOMETRY_NODE_VERSION="$(node -v 2> /dev/null)"
+            GEOMETRY_NPM_VERSION="$(npm --version 2> /dev/null)"
```
desyncr: Now, here is the thing. What we want to do is to:

1 - Check if we're in a node project
2 - Check which package manager the project is using (either `npm` or `yarn`)

With point 1) we are trying to avoid displaying node/npm/yarn version unless it's needed at all (ie, when the user is sitting in a node project).
We can achieve this by verifying the existence of `package.json`, which conventionally is used by `npm` and by `yarn`.  This check alone solves point 1):

    if [[ ! -f "$PWD/package.json" ]]; then # Check we're in a node project
        return # Don't print anything when we're not...
    fi

In regard of point 2) we'll need to check for a `yarn.lock` file and the existence of `yarn` command. Why? Well, a `yarn.lock` will tell us the project uses `yarn` rather than `npm` (both of them use `package.json`, so it doesn't tell us much). We also need to check for `yarn` command because the user may checkout a project that is using `yarn` but he may not have it installed yet (`npm` comes with `node` installation, so in this case there is no need to check it).
So, all in all it may look as follows:

    local packager=npm # Use npm by default (comes with node installation)
    if [[ $+commands[yarn] && -f "$PWD/yarn.lock" ]]; then # Check yarn and yarn.lock exists
        packager=yarn # If we're in a project using yarn and yarn command is available, use it instead of npm
    fi
    NPM_YARN_VERSION=$($packager --version 2> /dev/null) # Display either npm or yarn version

In the line above we're using `packager` variable (which is a string, either "npm" or "yarn") to dynamically form a command with one of these two strings, depending on the condition above, it may execute `npm --version 2> /dev/null` or `yarn --version 2> /dev/null`.

And that's all, we then will display `node` version and `npm/yarn` version.

It does look like a lot of things to do but we're almost there :)

plugins/node/plugin.zsh
castellanprime: I fixed that in the pull request.

desyncr (changes requested):
There're a couple of things to look for. I tried to be as descriptive as possible. Let me know if you've any question or need help with anything!

desyncr:
Btw we should implement similar behaviour to other project-based plugins, such as `ruby`, `rustup` and `virtual_env`. Only display these prompts when we're sitting in a project (`Gemfile`, `Cargo.toml`, etc.)

@frmendes What you think? I'll open another issue if you agree or want to discuss it further.

plugins/node/plugin.zsh
castellanprime: Should we differentiate between node and yarn in the prompt?

desyncr: We could add a `* yarn` thing, or show another icon.

frm:
I'll review this later, but regarding this comment:

> we should implement similar behaviour to other project-based plugins

I agree and I think we should encourage this more. I'm not a big fan of abstracting more on top of the plugin layer just yet (say, by providing utility functions to render language plugins if a given file is present). Maybe in the future, but for now definitely set some guidelines or similar.

plugins/node/plugin.zsh
castellanprime: I meant npm and yarn.

desyncr:
@castellanprime Looking really great!

plugins/node/plugin.zsh
```diff
@@ -1,18 +1,32 @@
...
-        echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
+    if [[ ! $+commands[node] ]]; then 
+        return
+    fi
+
+    if [[ ! -f "$PWD/package.json" ]]; then
+        return
     fi
+
+    if [[ $+commands[yarn] && -f "$PWD/yarn.lock" ]]; then
```
desyncr: Something I'm thinking now is that `GEOMETRY_NODE_DEFAULT_PACKAGE_MANAGER` won't be restored to `npm` after entering a `yarn` based project. So, we should add a `else` statement to `GEOMETRY_NODE_DEFAULT_PACKAGE_MANAGER="npm"`.

plugins/node/plugin.zsh
```diff
@@ -1,18 +1,32 @@
...
+        GEOMETRY_NODE_DEFAULT_PACKAGE_MANAGER="yarn"
+    fi 
+
+    GEOMETRY_PACKAGER_VERSION="$($GEOMETRY_NODE_DEFAULT_PACKAGE_MANAGER --version 2> /dev/null)" 
+    GEOMETRY_NODE_VERSION="$(node -v 2> /dev/null)"
+    echo "$GEOMETRY_NODE_PACKAGER_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_PACKAGER_VERSION)"
```
desyncr: We can have a mark here to show which version are we displaying:

    ($GEOMETRY_NODE_DEFAULT_PACKAGE_MANAGER:$GEOMETRY_PACKAGER_VERSION)

It will look kind of: `v6.9.5 (npm: 3.10.10)`. Just a suggestion.

desyncr (changes requested):
One small thing left :)

plugins/node/plugin.zsh
castellanprime: Okay.

desyncr (approved):
So approved. Great work!

Thanks a lot for the time invested and the awesome contributions to geometry!

Gonna approve now and merge later tonight :-)

castellanprime:
@desyncr @frmendes Thanks for letting me on. If you have any other suggestions on how to improve this, feel free to let me know. So can I take another issue and tackle it?

desyncr:
There're plenty of [open](https://github.com/frmendes/geometry/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22) issues feel free to join any of those :) If you got suggestions or ideas how to improve the project in anyway please open an issue or join us at [gitter](https://gitter.im/geometry-zsh/Lobby).

desyncr:
Looking and working great from my side. @frmendes Any feedback?

frm (approved):

frm:
Looking good, merging. Thanks 😄

PR #116, opened 2017-02-10
Add note contextual plugin rendering

desyncr:
Added a note about contextual plugin rendering.

### DIFF
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -84,8 +84,8 @@ geometry_prompt_pretty_git_render() {
 ```
 
 As a convention plugins should only be rendered when sitting on a valid context,
-per example the `node` built-in plugin will only be displayed when sitting on a
-node project. This is done in order to have a uncluttered prompt.
+for example the `node` built-in plugin will only be displayed when sitting on a
+npm/yarn-based project. This is done in order to have a uncluttered prompt.
 
 Finally you'll need to "register" the plugin with geometry in order to set it up
 and render it on each render cycle.
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -85,7 +85,8 @@ geometry_prompt_pretty_git_render() {
 
 As a convention plugins should only be rendered when sitting on a valid context,
 for example the `node` built-in plugin will only be displayed when sitting on a
-npm/yarn-based project. This is done in order to have a uncluttered prompt.
+npm/yarn-based project. This is done in order to have a uncluttered prompt, we
+encourage you to have this in mind.
 
 Finally you'll need to "register" the plugin with geometry in order to set it up
 and render it on each render cycle.
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -83,6 +83,10 @@ geometry_prompt_pretty_git_render() {
 }
 ```
 
+As a convention plugins should only be rendered when sitting on a valid context,
+per example the `node` built-in plugin will only be displayed when sitting on a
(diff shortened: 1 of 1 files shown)


jedahan:
I thought about this when creating the rustup plugin, and decided if it was enabled to always render. My reasoning was that I often will `cargo install`, or `cargo init` and want to know if i am on stable, beta, or nightly. I see a few options, I am open to all of them or any other suggestions.

* Only render if in a git repository with Cargo.toml
* Provide a GEOMETRY_RUSTUP_ALWAYS_RENDER env variable
* Refactor geometry to allow adding a * or something next to a plugin, to have it always render
* Look at the words on the prompt and if `cargo` is in it, then trigger prompt rerender (same for npm, etc...)

desyncr:
> Only render if in a git repository with Cargo.toml

I would not mix both of them. What If they use another scm, or none at all.

> Provide a GEOMETRY_RUSTUP_ALWAYS_RENDER env variable

I thought about this configuration as well.

> Refactor geometry to allow adding a * or something next to a plugin, to have it always render
> Look at the words on the prompt and if cargo is in it, then trigger prompt rerender (same for npm, etc...)

Not really sure about this. What's the intent?

jedahan:
> I would not mix both of them. What If they use another scm, or none at all.

Generally if I am in any subdirectory of something with a Cargo.toml, I'd like to keep an eye on the rustup version. But it is a good idea to separate concerns, so I'll probably remove the git check, and only look for *Cargo.toml* in the current directory.

>> Provide a GEOMETRY_RUSTUP_ALWAYS_RENDER env variable
> 
> I thought about this configuration as well.

I am leaning towards this - the idea being if we see enough plugins with a similar option, then we can decide to make it easier to write.

> Refactor geometry to allow adding a * or something next to a plugin, to have it always render
> 
> Not really sure about this. What's the intent?

I made a small hack proof-of-concept to illustrate what I mean better.

>> Look at the words on the prompt and if cargo is in it, then trigger prompt rerender (same for npm, etc...)
> 
> Not really sure about this. What's the intent?

> plugins should only be rendered when sitting on a valid context

I was daydreaming about 'what is a valid context', if I were to type the word 'cargo', I am about to do something with cargo, therefore I am in a context of knowing what rust toolchain is currently selected. It is quite outside the scope of regular geometry plugins.

plugins/README.md
```diff
@@ -83,6 +83,10 @@ geometry_prompt_pretty_git_render() {
 }
 ```
 
+As a convention plugins should only be rendered when sitting on a valid context,
+for example the `node` built-in plugin will only be displayed when sitting on a
+npm/yarn-based project. This is done in order to have a uncluttered prompt.
```
frm: I think we could be a bit more clear on what we want users to follow. Something like "the `node` built-in plugin will check if a `package.json` or `yarn.lock` file is present before rendering. This means it will only be displayed when sitting (...). When developing a custom plugin, we encourage you to have this in mind."

Or equivalent.

frm (changes requested):

frm (approved):

frm:
Given we now have #117 , I think we can merge this. @jedahan what's your take?

jedahan:
I am rewriting this @frmendes and will have a new pull request by the end of the day (EST)

jedahan:
@frmendes I think this is still a nice paragraph to add.

desyncr:
Cool. Merging. 👍

PR #117, opened 2017-02-10
Initial proposal for 'starred' plugins

jedahan:
A star* can mean anything, usually 'always display this plugin'

It will be passed in as the first parameter to the setup()

I am open to other ideas, like

* always passing it in as the first arg to render()
* making it a different symbol than *, since that requires quoting
* creating a helper function to check if a plugin is starred

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ fast™.
 
 ## Plugins
 
-geometry has an internal plugin architecture. The default plugins are `exec_time`, `git` and `hg`. But you can enable a variety of built-in plugins just by setting the `GEOMETRY_PROMPT_PLUGINS` variable in your own configuration files:
+geometry has an internal plugin architecture. The default plugins are `exec_time`, `git` and `hg`.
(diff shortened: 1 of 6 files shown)


frm:
I like the idea of having a plugin always render. For example, I use a global version of ruby but then have a lot of project-specific versions. I'm not 100% certain on this approach and would like feedback.

Just a small note: when I read "starring", my mind initially went to "favoriting". What do you think about using something like "pinning" to avoid ambiguity?

jedahan:
"pinning" is a good word, if we are ok with making that be the only functionality.

If we say that * = pin = always display, then maybe plugins can define a third function:

`setup()`, `check()`, `render()`, where geometry skips the `check()` function if pinned.

This removes the need for having custom variables for each plugin. I will create a new pull request for "plugin-pinning"

desyncr:
I like the `check` method, it'll simplify current `render` method, making it overall easier to understand and contribute new plugins as well.

frm:
Great idea! I really like that alternative! 👍

jedahan:
See #119

frm:
@jedahan can we delete this branch or do you still have something in it?

PR #119, opened 2017-02-10
lib/plugin: support plugin pinning

jedahan:
This adds geometry_prompt_plugin_check(), which will be run to determine if we are in a context where it makes sense to display the plugin

A plugin can be pinned by appending a * to skip this check.

How can we could remove the quoting requirement for pins?

    GEOMETRY_PROMPT_PLUGINS+=(rustup*)

instead of

    GEOMETRY_PROMPT_PLUGINS+=('rustup*')

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ documentation and configuration for each specific plugin in the
 [plugins](/plugins) directory.
 
 Some plugins only render when you are in a given directory or in the presence of a given file.
-You can have those plugins always render by pinning a `+` before the name
+You can have those plugins always render by pinning a `+` before the name.
 
 ```sh
 export GEOMETRY_PROMPT_PLUGINS=(exec_time git +rustup) # rustup will always render
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ fast™.
 
 ## Plugins
 
-geometry has an internal plugin architecture. The default plugins are `exec_time`, `git` and `hg`. But you can enable a variety of built-in plugins just by setting the `GEOMETRY_PROMPT_PLUGINS` variable in your own configuration files:
+geometry has an internal plugin architecture. The default plugins are `exec_time`, `git` and `hg`.
+But you can enable a variety of built-in plugins just by setting the `GEOMETRY_PROMPT_PLUGINS` variable in your own configuration files:
 
 ```sh
 GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg)
@@ -74,6 +75,12 @@ GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg)
 These plugins will load and display on the right prompt. You can check the
 documentation and configuration for each specific plugin in the
 [plugins](/plugins) directory.
+
+You can pin* a plugin by appending `*` to the end of the name. Geometry will always render the plugin.
+
+```sh
+GEOMETRY_PROMPT_PLUGINS+=('rustup*')
+```
 
 geometry also supports your own custom plugins. See the plugin [documentation](/plugins/README.md) for
 instructions and examples.
--- a/README.md
+++ b/README.md
@@ -76,10 +76,11 @@ These plugins will load and display on the right prompt. You can check the
 documentation and configuration for each specific plugin in the
 [plugins](/plugins) directory.
 
-You can pin* a plugin by appending `*` to the end of the name. Geometry will always render the plugin.
+Some plugins only render when you are in a given directory or in the presence of a given file.
+You can have those plugins always render by pinning a `+` before the name
 
 ```sh
-GEOMETRY_PROMPT_PLUGINS+=('rustup*')
+export GEOMETRY_PROMPT_PLUGINS=(exec_time git +rustup) # rustup will always render
 ```
 
 geometry also supports your own custom plugins. See the plugin [documentation](/plugins/README.md) for
--- a/plugins/README.md
+++ b/plugins/README.md
@@ -23,10 +23,12 @@ configuration files.
 
 
 ```sh
-GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg 'rustup*')
+GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg +rustup)
 ```
 
 *Note: if you're not sure where to put geometry configs, just add them to your `.zshrc`*
+
+*Note: the `+` before rustup means the plugin is pinned, and will always render, regardless of context*
 
 ## Custom plugins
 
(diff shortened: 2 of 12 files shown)


jedahan:
### Some alternatives to `*` at the end

1. `plugin.` looks like a typo to me

```sh
export GEOMETRY_PROMPT_PLUGINS=(exec_time git node rustup.)
```

2. `.plugin` look like a pin, and less like a typo

```sh
export GEOMETRY_PROMPT_PLUGINS=(exec_time git node .rustup)
```

3. `plugin!` looks alarming
```sh
export GEOMETRY_PROMPT_PLUGINS=(exec_time git node rustup!)
```

desyncr (commented):
I have no issues with quoting. Using '*' is cool to me.

desyncr:
Watch out for API breaking changes. Either we should be backward compatible or start tagging/versioning.

My concern is with existing custom plugins that might break with a geometry update.

jedahan:
One thing I saw zplug do, which I like, is to release a version that prints out deprecation messages, then wait a month, then remove support. We can do that here, though we might be able to get away with just ignoring `_check()` if it does not exist, which would be backwards compatible.

README.md
```diff
@@ -75,6 +76,12 @@ These plugins will load and display on the right prompt. You can check the
 documentation and configuration for each specific plugin in the
 [plugins](/plugins) directory.
 
+You can pin* a plugin by appending `*` to the end of the name. Geometry will always render the plugin.
```
frm: I would add a note here on pinning, since when reading this document from the start you don't really get what it is. If I was a new user, I would think "wait, so plugins don't always render? Why not?"

What do you think of just prepending something like "Some plugins only render when you are in a given directory or in the presence of a given file. You can have them always render by pinning the plugin. (...)"?

plugins/README.md
```diff
@@ -23,15 +23,15 @@ configuration files.
 
 
 ```sh
-GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg)
+GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg 'rustup*')
 ```
 
 *Note: if you're not sure where to put geometry configs, just add them to your `.zshrc`*
```
frm: For the same reasons as before, if I missed the pinning section, I would be confused with the `'rustup*'`. Maybe add a link saying "if you're not sure what's with the `*` you can read about it [here]()" ? What's your take?

plugins/README.md
```diff
@@ -62,19 +62,25 @@ for now, so let's leave it blank.
...
+Now, checking. The `check` function is called before `render`, and should check if
+it makes sense to display the plugin in the current context, returning non-zero if
+we should skip rendering.
+
+```sh
+geometry_prompt_pretty_git_check() {
+  # Do nothing if we're not in a repository
+  [ -d $PWD/.git ] || return 1
+}
+```
```
frm, line 79: 👍 nice example, clear explanation

plugins/rustup/README.md
```diff
@@ -1,13 +1,18 @@
...
 ## Installation
 
     cargo install rustup_prompt_helper
 
 ## Configuration
 
+```sh
+GEOMETRY_PROMPT_PLUGINS+=('rustup*')
+```
+
```
frm: Note about pinning here? Since it is not required the plugin with the *

plugins/node/plugin.zsh
```diff
@@ -5,14 +5,18 @@ GEOMETRY_COLOR_NODE_NPM_VERSION=${GEOMETRY_COLOR_NODE_NPM_VERSION:-black}
...
+    test -f package.json || return 1
+}
 
 geometry_prompt_node_render() {
-    if (( $+commands[node] )); then
-        GEOMETRY_NODE_VERSION="$(node -v 2> /dev/null)"
-        GEOMETRY_NPM_VERSION="$(npm --version 2> /dev/null)" 
-        echo "$GEOMETRY_NODE_NPM_VERSION $GEOMETRY_NODE_VERSION ($GEOMETRY_NPM_VERSION)"
-    fi
+    GEOMETRY_NODE_VERSION="$(node -v 2> /dev/null)"
```
frm: Just a note: don't forget to update this with the code recently merged from #108

frm (changes requested):
Overall looks pretty good to me. Just some notes on the reads and a note on updating the node plugin.

Thanks for taking the time to add checks to the current plugins!

frm:
I'm good with 'plugin*', don't have any strong feelings on that. None of the other suggestions do it better IMO.

I moved the version discussion to #121 since a lot has been said in different PRs and issues. Might as well have a dedicated issue for that 😄

plugins/rustup/README.md
jedahan: removed the pinning

plugins/README.md
jedahan: Added  a note beneath the current note

README.md
jedahan: Good idea, prepended a sentence about pinning

plugins/node/plugin.zsh
jedahan: Rebased against master, will force push to this PR soon

jedahan:
I really don't like the extra ritual for globbing, this patch proposes `+plugin` for pinning

jedahan:
I also made it backwards compatible in that, if you don't have a check function for a plugin, it just runs render. That way we don't need to version yet.

frm (approved):

frm:
Looking great! @desyncr any takes?

README.md
```diff
@@ -75,6 +76,13 @@ These plugins will load and display on the right prompt. You can check the
 documentation and configuration for each specific plugin in the
 [plugins](/plugins) directory.
 
+Some plugins only render when you are in a given directory or in the presence of a given file.
+You can have those plugins always render by pinning a `+` before the name
```
desyncr: "before the name" => "before the name:"

plugins/README.md
```diff
@@ -23,15 +23,17 @@ configuration files.
 
 
 ```sh
-GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg)
+GEOMETRY_PROMPT_PLUGINS=(virtualenv docker_machine exec_time git hg +rustup)
 ```
 
 *Note: if you're not sure where to put geometry configs, just add them to your `.zshrc`*
 
+*Note: the `+` before rustup means the plugin is pinned, and will always render, regardless of context*
```
desyncr: "pinned" should like to the "Pinning" section.

desyncr (commented):
Two small things. Should be good to go 👍

jedahan:
So, should I squash the three commits, rewrite the commit message, and push? Not sure what's good protocol.

frm:
Precisely that. Squash, merge, push. Or just change the big green button on GitHub to say "Squash and merge" and hit that.

desyncr (approved):

PR #126, opened 2017-02-20
Fix virtualenv plugin checking condition

### LINKED ISSUE #125: Virtualenv plugin error
With the `virtualenv` plugin enabled:

```
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]

 ▲ ~/Developer/dotfiles                                         () master :: 91d :: •
```

I recently pulled off master so it was probably introduced since 414ff89, but I can't double check it right now.

desyncr:
Fixes #125 

Fixes this specific case. I still have to review other `_check` implementations.


jedahan:
I think this is equivalent to `test $VIRTUAL_ENV || return 1`

jedahan (approved):

PR #127, opened 2017-02-25
Fix rprompt plugin separator rendering

jedahan:
This fixes a bug I had with trailing plugins whose check would fail


frm:
Wait, what bug is this referring to?

jedahan:
If the last plugin `_check` returns 1, and therefore should not render, a trailing separator will still be in rprompt because it was appended by the previous plugin.

desyncr:
Looking fine!

PR #128, opened 2017-02-25
Initial hydrate plugin

jedahan:
This shows a little droplet if you should drink water, which can be removed by running `hydrate`

### DIFF
--- a/plugins/hydrate/README.md
+++ b/plugins/hydrate/README.md
@@ -1,6 +1,6 @@
 # Hydrate
 
-Reminds you to hydrate every once in a while.
+Reminds you to hydrate every once in a while. Run `hydrate` once you have hydrated.
 
 ## Installation
 
--- a/plugins/hydrate/README.md
+++ b/plugins/hydrate/README.md
@@ -1,0 +1,17 @@
+# Hydrate
+
+Reminds you to hydrate every once in a while.
+
+## Installation
+
+```sh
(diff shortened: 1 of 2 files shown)


frm:
This is great! I'm really into this kind of healthy promotion. I'm up for merging but can we actually improve this into a `health` plugin? Stretching, standing up in an interval, etc.

desyncr:
I believe this should go into https://github.com/frmendes/geometry/wiki/Plugins.

jedahan:
Added to the plugins page for now

jedahan:
@frmendes how about I allow overriding the command to dismiss the icon? so you can do something like GEOMETRY_PLUGIN_HYDRATE_COMMAND="stretch" or something like that?

frm:
Well, any command is fine.

What's been bothering is that I'd just like to set a bunch of indicators. Hydrate, stretch, break... Lately I've been sprinting with the pomodoro timer technique and aprox every 25 minutes I just stop. And then run `restart` after my break, for another 25 minutes. This is different from hydrate because, well, I just want a bunch of reminders.

I wonder if it's possible to cram up a bunch of symbols and timers into an array and display those symbols after the corresponding timer has gone through.

Or maybe we're overkilling it and I should just restrain myself to my pomodoro app.

desyncr:
Should I close this PR as it's in the Wiki.

jedahan:
Yeah feel free to close

PR #129, opened 2017-02-26
Fix prefix check to allow empty string

### LINKED ISSUE #113: Geometry handle empty prefix
Users cannot configure `GEOMETRY_PROMPT_PREFIX`  as "" (empty string) due a faulty check [here](https://github.com/frmendes/geometry/blob/master/geometry.zsh#L36).

See https://github.com/frmendes/geometry/issues/112

kguthrie:
Allows user to specify GEOMETRY_PROMPT_PREFIX="" for single line usage

Fixes #113


desyncr:
Thanks! I'm gonna take a look today!

desyncr:
I'm afraid it shows the same behavior. Can you share your test case with configuration?

desyncr (changes requested):
It seems it shows the current behavior.

kguthrie:
I'm a bit new at all this...sorry! I tested it out yesterday and thought it was working, ~~but now that I give it a try this morning I agree with you, it doesn't seem to work~~. I'll get back to you.

kguthrie:
Maybe I got ahead of myself. Here is what I see right now:

image

desyncr:
My bad. It look great! Thanks for the contribution!

desyncr (approved):

PR #130, opened 2017-03-05
plugin/git: fix errors inside bare repo

PlanteVodu:
Hi there!

Sometimes you need to go into a bare repository, and then you get an error:

error

As you see, the git plugin is executing commands that it should not run here.

I tried to fix that. Now it displays a symbol instead, indicating you're in a bare repo:

fixed

Color & symbol can be set by the user.

Please tell me if you have any suggestions :)


plugins/git/plugin.zsh
```diff
@@ -150,6 +153,12 @@ geometry_prompt_git_check() {
 }
 
 geometry_prompt_git_render() {
+  # Check if we are in a bare repo
+  if [[ $(command git rev-parse --is-inside-work-tree 2>/dev/null) != "true" ]] ; then
```
frm: A bit of nitpicking but this `!= "true"` is kind of itching me. According to the docs, it returns `"false"` otherwise. Can we test for that?

frm (changes requested):
It looks pretty good! Thanks! Just that one comment.

plugins/git/plugin.zsh
FildasPA: Is it okay like this?

PlanteVodu:
Is it suitable like this?

frm (approved):
I was using this for the day. Seems to be good. Thanks!

PR #132, opened 2017-03-14
Support plugin pinning with built-in plugins

### LINKED ISSUE #131: Allow sourcing plugins from any directory
Right now, we default to `$GEOMETRY_ROOT/plugins/$plugin/plugin.zsh`

If we detect that `$plugin` has any `/` in it, maybe we should source that plugin as a full path?

This would help to not need people to clone plugins into `$GEOMETRY_ROOT/plugins/`

desyncr:
- Support plugin pinning for custom and builtin plugins
- Moved pinning logic to register method

The way a built-in plugin self-register as pinning is by calling `geometry_plugin_register +plugin-name`.

As a future TODO:

- [ ] Support `pin` and `unpin` methods

Fixes #131

### DIFF
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -16,11 +16,6 @@ typeset -ga _GEOMETRY_PROMPT_PLUGINS_PINNED
 # Set up default plugins
 geometry_plugin_setup() {
   for plugin in $GEOMETRY_PROMPT_PLUGINS; do
-    if [[ $plugin[1] == '+' ]]; then
-      plugin=${plugin#?}
-      _GEOMETRY_PROMPT_PLUGINS_PINNED+=$plugin
-    fi
-
     source "$GEOMETRY_ROOT/plugins/$plugin/plugin.zsh"
   done
 }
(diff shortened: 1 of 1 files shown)


frm (changes requested):
Looks great! But isn't the commit message wrong? Shouldn't it be "custom plugins"?

lib/plugin.zsh
```diff
@@ -39,6 +34,11 @@ geometry_plugin_register() {
     return 1
   fi
```
jedahan, line 35: Won't this always return false for pinned plugins? Here we check against `+plugin` but only ever register `plugin` in `_GEOMETRY_PROMPT_PLUGINS`

jedahan (changes requested):
I think there is a way to get rid of the _GEOMETRY_PROMPT_PLUGINS_PINNED array completely, simplifying the code a bit.

jedahan:
So the ergonomics for someone to pin a custom plugin, with this patchset:

```zsh
source $HOME/.zpm-init.zsh || {
  zpm frmendes/geometry
  zpm jedahan/geometry-hydrate
  exec zsh
}
geometry_plugin_unregister hydrate
geometry_plugin_register +hydrate
```

I could remove the unregister requirement from geometry-hydrate, but this would make it diverge more from main plugins because all the main plugins self register.

desyncr:
@jedahan The proposed approach gives the power to decide to be pinned or not to the plugins themselves. In this way, if you want `geometry-hydrate` plugin to be always pinned, you'll need to adapt the plugin to register this way:

     geometry_plugin_register +hydrate

If you think this plugin may either be pinned or not this plugin should support a configuration variable, also implemented by the plugin.

What do you think about this approach? Given you developed the pinning functionality you have a better idea of what should and should not go into it.

desyncr:
I believe your approach in #133 is more complete that this one. Let's move onto that!

PR #133, opened 2017-03-15
Simplify plugin pinning

### LINKED ISSUE #131: Allow sourcing plugins from any directory

jedahan:
Instead of keeping a separate array around, we just check
if a plugin is pinned in GEOMETRY_PROMPT_PLUGINS.

This allows third party plugins to be sourced anywhere,
registered when they want, and runtime check for pins.

This also makes it easier to runtime (un)pin.

Fixes #131 and #132

### DIFF
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -10,18 +10,10 @@ fi
 # List of active plugins
 typeset -ga _GEOMETRY_PROMPT_PLUGINS
 
-# List of pinned plugins
-typeset -ga _GEOMETRY_PROMPT_PLUGINS_PINNED
-
 # Set up default plugins
 geometry_plugin_setup() {
   for plugin in $GEOMETRY_PROMPT_PLUGINS; do
-    if [[ $plugin[1] == '+' ]]; then
-      plugin=${plugin#?}
-      _GEOMETRY_PROMPT_PLUGINS_PINNED+=$plugin
-    fi
-
-    source "$GEOMETRY_ROOT/plugins/$plugin/plugin.zsh"
+    test -f "$GEOMETRY_ROOT/plugins/${plugin#+}/plugin.zsh" && source $_
   done
 }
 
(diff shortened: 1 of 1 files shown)


lib/plugin.zsh
```diff
@@ -104,3 +92,43 @@ geometry_plugin_render() {
 
   echo "$rprompt"
 }
+
+geometry_plugin_pin() {
+  if [[ $# -eq 0 ]]; then
+    echo "Error: Missing argument."
```
frm: Can you change this for something like "Missing the plugin name"? Just to be a tad bit more clear for when users experiment with this.

lib/plugin.zsh
```diff
@@ -104,3 +92,43 @@ geometry_plugin_render() {
...
+    GEOMETRY_PROMPT_PLUGINS[$GEOMETRY_PROMPT_PLUGINS[(i)$plugin]]=(+$plugin)
+    return 0
+  fi
+
+  GEOMETRY_PROMPT_PLUGINS+="+$plugin"
+}
+
+geometry_plugin_unpin() {
+  if [[ $# -eq 0 ]]; then
+    echo "Error: Missing argument."
```
frm: Same as previous comment.

frm (changes requested):
Well, :clap: :clap: :clap: for simplifying this to a single array.

As for runtime pinning/unpinning it can be useful. I don't see why not even though I haven't felt this need myself. This usually makes me reconsider whether we need it, but the increase in complexity isn't that much, it's not a bad trade-off.

However, I'd at least add a note for unaware users that the pinning is limited to the current session. Once that goes off, default settings get re-applied.

jedahan:
@frmendes I prefer to only apply the first commit, and keep things simple. I created `geometry_plugin_(un)pin` since it was in the TODO for the previous pull request, as a proof of concept.

frm:
@jedahan precisely that. I haven't felt the need to add those functions but they **can** be useful. Are they, though? The increase in complexity is not too big in this case since it's just a couple of `if`s although it may come back to bite us if we ever decide to change the actual plugin implementation.

that first commit, though ❤️

jedahan:
It seems we are in agreement regarding the first commit, mind if I close this PR and push it?

frm (approved):
Go right on ahead and merge it.

Issue #63, opened 2016-12-14
Aliased commands may break geometry


desyncr:
As per https://github.com/frmendes/geometry/issues/59. Aliased commands may break/change geometry inner workings.

Avoid this with a similar approach to https://github.com/zsh-users/zsh-syntax-highlighting/issues/390:

At the top:

    # Save aliases into a variable and remove/unalias everything
    zsh_highlight__aliases=`alias -L`
    builtin unalias -m '*'

At the bottom:

    # Restore aliases & kill global variable
    eval "$zsh_highlight__aliases"
    builtin unset zsh_highlight__aliases

frm:
As I mentioned in #59 looking through the code there don't seem to be many instances of possible aliased commands.

A possible solution would be prepending with `command rm` or `\rm`, even though that makes the code messy.

Unaliasing and aliasing over and over is going to affect render time, isn't it?

desyncr:
I guess this issue was left hanging out since this issue was resolved with `\rm`.

PR #134, opened 2017-03-16
Moves the cleanup of the async tempfile from on-startup to on-exit

### LINKED ISSUE #114: Async library tries to erase temporal files from other users
[Async lib](https://github.com/frmendes/geometry/blob/master/lib/async.zsh) uses temporal files to communicate with background jobs.

When a new prompt is loaded the function `geometry_async_setup` tries to clean up -possibly- stalled temporal files. In doing so it does not check the ownership (ex, a prompt for `root`). So it may face a prompt to confirm deletion (`zsh 5.3`).

A possible solution is to use `$UID` to [prefix temporal files](https://github.com/frmendes/geometry/blob/master/lib/async.zsh#L3):

    typeset -g GEOMETRY_ASYNC_TMP_FILENAME_PREFIX='.geometry-rprompt-info-'$UID'-'

thomasdesr:
The core issue in #114 is that if `rm` hits a file it couldn't delete (generally due to some form of  permission issues), it will prompt the user.

Where this generally helpful behavior becomes problematic is when async updates are enabled. `geometry_async_setup` currently attempts to clean up any old or leaked async temp files by running `rm` on them at startup. When the prompt triggers it ends up blocking the startup of the zsh shell since `geometry_async_setup` is in the hotpath for shell startup.

There were three options discussed for fixing this:
* Add the UID to tempfile prefix so the startup cleanup could avoid the discovered cause of permission issues (sudoed shells creating root owned files that the normal user can't delete)
* Make the deletion forceful, this will prevent `rm` from prompting
* Add an on-exit hook that will clean up the specific tempfile used by that zsh process

This PR implements the 3rd option.

(Fixes #114)

### DIFF
--- a/lib/async.zsh
+++ b/lib/async.zsh
@@ -23,12 +23,17 @@ typeset -g GEOMETRY_ASYNC_PROC_ID=0
     GEOMETRY_ASYNC_PROC_ID=$!
 }
 
-# Hooks precmd to run `-geometry_async_function` and hooks
-# `USR1` signal to comunicate with it
+# Removes the async temp file when zsh exits
+-geometry_async_zshexit() {
+    rm -f "${GEOMETRY_ASYNC_TMP_FULL_PATH}$$"
+}
+
+# geometry_async_setup adds zsh-hooks into precmd and zshexit, as well as a
+# signal trap on SIGUSR1, to support communication of asynchronous updates to
+# the running terminal.
 geometry_async_setup() {
     add-zsh-hook precmd '-geometry_async_precmd'
-
-    (\rm ${GEOMETRY_ASYNC_TMP_FULL_PATH}*) &> /dev/null
+    add-zsh-hook zshexit '-geometry_async_zshexit'
 
     TRAPUSR1() {
         # read from temp file
--- a/lib/async.zsh
+++ b/lib/async.zsh
@@ -25,7 +25,7 @@ typeset -g GEOMETRY_ASYNC_PROC_ID=0
 
 # Removes the async temp file when zsh exits
 -geometry_async_zshexit() {
-    rm -f "${GEOMETRY_ASYNC_TMP_FULL_PATH}$$"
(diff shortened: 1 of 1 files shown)


thomasdesr:
cc @frmendes @desyncr

desyncr (approved):
Looking and working pretty fine! Thanks for the work!

lib/async.zsh
```diff
@@ -23,12 +23,17 @@ typeset -g GEOMETRY_ASYNC_PROC_ID=0
     GEOMETRY_ASYNC_PROC_ID=$!
 }
 
-# Hooks precmd to run `-geometry_async_function` and hooks
-# `USR1` signal to comunicate with it
+# Removes the async temp file when zsh exits
+-geometry_async_zshexit() {
+    rm -f "${GEOMETRY_ASYNC_TMP_FULL_PATH}$$"
```
jedahan: Minor but maybe we want to prefix the `rm` with `\` or `builtin` ?

jedahan (changes requested):

lib/async.zsh
frm: I'm with @jedahan on this one. We should force the usage of the actual `rm` as there were similar problems in #63.

Correct me if I'm wrong but you dropped the `/dev/null` redirection since it's now a hook, correct?

frm (changes requested):

lib/async.zsh
thomasdesr: `\` is done

I did remove the redirection, but that wasn't intentional. I can add it back if you believe we shouldn't report this error to the user.

jedahan (approved):

lib/async.zsh
frm: I kinda want to, to avoid users mistaking it for an issue, but I feel like that might potentially obscure bugs. I'll approve and merge it. Thanks!

frm (approved):

PR #135, opened 2017-03-16
Initial background-jobs plugin implementation

### LINKED ISSUE #110: Background jobs support
Is there any plan to add some sort of indicator for background jobs?

desyncr:
Bare bones background-jobs plugin. No icon, color or configuration available. All suggestions welcome.

It will display as follows if there is one or more background jobs:

screenshot from 2017-03-15 22-54-10

It won't display anything if there is no background job.

Fixes #110

### DIFF
--- a/plugins/jobs/plugin.zsh
+++ b/plugins/jobs/plugin.zsh
@@ -11,7 +11,8 @@ geometry_prompt_jobs_check() {
 }
 
 geometry_prompt_jobs_render() {
-  echo $(prompt_geometry_colorize $GEOMETRY_COLOR_JOBS '%(1j.${GEOMETRY_SYMBOL_JOBS} %j.)')
+  local jobs_prompt='%(1j.'$GEOMETRY_SYMBOL_JOBS' %j.)'
+  echo $(prompt_geometry_colorize $GEOMETRY_COLOR_JOBS $jobs_prompt)
 }
 
 geometry_plugin_register jobs
--- a/plugins/jobs/plugin.zsh
+++ b/plugins/jobs/plugin.zsh
@@ -4,7 +4,9 @@ GEOMETRY_COLOR_JOBS=${GEOMETRY_COLOR_JOBS:-blue}
 # Symbol definitions
 GEOMETRY_SYMBOL_JOBS=${GEOMETRY_SYMBOL_JOBS:-"⚙"}
 
-geometry_prompt_jobs_setup() {
+geometry_prompt_jobs_setup() {}
+
+geometry_prompt_jobs_check() {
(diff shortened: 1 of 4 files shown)


plugins/jobs/plugin.zsh
```diff
@@ -0,0 +1,14 @@
+geometry_prompt_jobs_setup() {
+  return true
```
jedahan: I like true/false better than 0/1, we should start using this convention in more places

plugins/jobs/plugin.zsh
```diff
@@ -0,0 +1,14 @@
...
+}
+
+geometry_prompt_jobs_check() {
+  return true
+}
```
jedahan: this is unnecessary - if the check function is not defined, it defaults to return true. If you like it for consistency or as an example its fine to keep.

jedahan (approved):
Looks like a solid start to me.

plugins/jobs/plugin.zsh
frm: TIL

plugins/jobs/plugin.zsh
frm: I agree. There are a number of rules we are starting to define that maybe would justify a small `CONTRIBUTING.md` file. What do you think?

frm (changes requested):
I like it but I'd rather have some colors and configurations already set up. It's not that much of a hassle that would justify an incremental build since you'd basically only need to add a color variable and a symbol variable or something like that.

plugins/jobs/plugin.zsh
jedahan: splitting out the readme into user-focused, and contributing to dev-focused information seems good to me.

desyncr: Agreed. Gonna create an issue to keep track of it.

plugins/jobs/plugin.zsh
desyncr: Yep, I wasn't sure about it so I left it after some debugging.

desyncr:
Gonna add color, symbol and if I can also a format string.

desyncr:
Sorry for the delay. Gonna give it a second iteration tonight. :)

desyncr:
screenshot from 2017-04-06 22-49-36

desyncr:
@frmendes Should be done now!

frm:
Sorry, Easter is a bit of a hassle here in Portugal. I'll get back to this shortly.

desyncr:
No problem!

frm (approved):

desyncr:
It's done!

PR #138, opened 2017-03-18
Fix malicious branch name shell injection

### LINKED ISSUE #137: Git plugin doesn't escape branch names
https://github.com/njhartwell/pw3nage means we are vulnerable

jedahan:
Described in https://github.com/njhartwell/pw3nage

If a branch name has something like `$(./pw3nage)`
we would execute it. Now we escape the $ symbol.

Fixes #137

### DIFF
--- a/plugins/git/plugin.zsh
+++ b/plugins/git/plugin.zsh
@@ -54,7 +54,7 @@ prompt_geometry_git_time_since_commit() {
 prompt_geometry_git_branch() {
   ref=$(git symbolic-ref --short HEAD 2> /dev/null) || \
   ref=$(git rev-parse --short HEAD 2> /dev/null) || return
-  echo "$(prompt_geometry_colorize $GEOMETRY_COLOR_GIT_BRANCH $ref)"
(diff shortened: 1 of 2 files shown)


frm (changes requested):
I was messing around with this yesterday and this solution doesn't cover cases where the branch name is ./pw3n with backticks surrounding

jedahan:
Yeah its a trickier problem than I thought.

So we want to disable shell expansion for the second parameter until the time it is actually rendered in the prompt. I will have to take a break and think about it more. If you come up with a good solution feel free to take it over.

frm:
Well, your solution is pretty much the same I had. I found it trickier to replace the backticks, so I just said "I'll think about it tomorrow". And here we are. I'll let you know if I come up with something.

frm:
Also, I'd prefer to avoid fixes like the one presented in https://github.com/njhartwell/pw3nage/pull/2 where all non-whitelisted characters are replaced with a `-`.

jedahan:
Closing in favour of #142

PR #139, opened 2017-03-18
Use return true/false instead of 0/1

jedahan:
Its feels nicer


frm (approved):

desyncr (approved):
Looks good! Merging right away!

PR #140, opened 2017-03-21
Revert "Use return true/false instead of 0/1"

desyncr:
Gonna revert this as I'm seeing some weird issues.

Reverts frmendes/geometry#139


frm:
Can you describe the issues?

desyncr:
It seems `true/false` aren't fully equivalent to `0/1`, I wanted to take a further look today but I couldn't. Plugin loading was broken (all plugins always load) -- so I reverted it.

jedahan:
Sorry for not testing thoroughly!

Issue #121, opened 2017-02-12
Version tagging


frm:
Continuing the conversation from #119 :

I think it might be about time to start tagging versions. Then use that to add a deprecation warning like @jedahan said. This means we should start adding a `CHANGELOG` and keep a list of deprecation dates in there.

What do you think of that option?

desyncr:
Deprecation messages at least for now are not that important IMO.

I would start with a minimal `CHANGELOG.md`, and `v0.1.0` then move from that. As of now I think we can go for it.

  - [ ] Create `CHANGELOG.md` with `v0.1.0` entry, just "Initial release" would suffice 
  - [ ] Update README.md with a release badge (see https://img.shields.io/github/release/qubyte/rubidium.svg)
  - [ ] Create release draft in https://github.com/frmendes/geometry/releases, thank our active contributors (castellanprime, ChauTNguyen) and link to https://github.com/frmendes/geometry/graphs/contributors

A future possibility is to have release schedules. But given our codebase and activity it seems unnecessary.

We may automate this but for now this is all we need to release `v0.1.0`.

Thoughts? @frmendes @jedahan

frm:
I'd like to take this on. Start with a minimal version and a `CHANGELOG` and build it up from there.

desyncr:
@frmendes Please do! I'm up for any question.

frm:
@jedahan @desyncr any features you would like to be built upon the implementation of version tagging? We already discussed deprecation messages.

jedahan:
Yeah I think the smallest subset of features for now, until we need them is good. We can probably put all message logic in the setup function on startup for people for now.

desyncr:
@frmendes Is there anything missing to implement versioning? Can I help with anything?

frm:
Well, time for me, tbh. Though, I promise I'll add it tonight 🤞

frm:
Fixed by #146

PR #146, opened 2017-04-05
Add changelog and tag v1.0.0

frm:
The project seems mature enough to mark it as `v1.0.0` but I'd actually prefer feedback on this.

I took the liberty of using semver and adding the [Keep A Changelog](http://keepachangelog.com/en/0.3.0/) format changelog.

After merging, I'll create the release draft and we can take it from there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,17 @@
+# geometry Change Log
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased]
+### Added
+
+### Changed
+
+## 1.0.0 - 2017-04-05
+### Added
+- Change Log file
+- Initial release features
+
+[Unreleased]: https://github.com/frmendes/geometry/compare/v1.0.0...HEAD
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # geometry
 
 [![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/geometry-zsh/Lobby)
+[![GitHub release](https://img.shields.io/github/release/frmendes/geometry.svg)](https://github.com/frmendes/geometry/releases/latest)
 
 geometry is a minimalistic, fully customizable zsh prompt theme.
 


frm:
It also fixes #121

desyncr:
I also believe `1.0.0` is a good starting point. Project is quite stable right now and won't face huge changes in the short term.

desyncr (approved):
I haven't used `Unreleased` section before but I have no issues against it, seems reasonable.

desyncr:
After @jedahan OK I think we should merge it and release right away. No need to wait for pending issues/PR.

What you think @frmendes?

CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased]
+### Added
+
+### Changed
+
+## 1.0.0 - 2017-04-05
```
jedahan, line 12: maybe we want to link to the current revision here?

jedahan (changes requested):
Good context, just want to add one link then we can push

CHANGELOG.md
jedahan: Or is that impossible because we have not pushed it yet?

frm: We have not pushed it yet. I could probably link to the revision and guess the correct URL. However, [Keep A Changelog](http://keepachangelog.com/en/0.3.0/) suggests linking the comparison between 2 consecutive versions.

frm:
@desyncr 

> I haven't used `Unreleased` section before but I have no issues against it, seems reasonable.

I haven't either but I'm willing to give it a shot and try that format. If you have any suggestions I'd love to hear them!

> I think we should merge it and release right away. No need to wait for pending issues/PR.

I agree 😄 as soon as @jedahan approves it, I'm releasing. Should I just make a release draft or can I release it right away as "Initial release"?

jedahan (approved):

jedahan:
Initial release is good. I wonder if we should do a little coordinated marketing around the release. Posting screenshots on r/unixporn, twitter, etc...

frm:
I agree with the marketing. Do you want to talk about this in an issue or gitter? @jedahan @desyncr

desyncr:
We should probably hang out on gitter to talk about it (there will be too much back and forward). We should set a date/time we can all hang out a couple of minutes.

frm:
@desyncr [Sure, sounds great!](https://youtu.be/GLbzN1Q1Agw?t=34s)

cc @jedahan

jedahan:
I am actually on vacation out of my normal country so it is a bit difficult for me to communicate but I will try. On April 19 I will be more reliable.

PR #158, opened 2017-05-18
Init git submodules when zsh-async not found

desyncr:
Added an error message when `zsh-async` library is not loaded. Not sure it's the best way. We may as well perform `git submodule update --init` ourselves.


frm (approved):
I'm approving this if you prefer to merge but I'd rather run the commands ourselves.

desyncr:
Updated PR.

frm (approved):

PR #160, opened 2017-05-21
zsh-async worker missing env variables

desyncr:
There was an issue with the zsh-async implementation which prevented registering custom plugins (see https://github.com/desyncr/geometry-dir-info-prompt/issues/4#issuecomment-302928204).

The issue was that we were creating a worker on the setup phase to submit jobs. The worker environment later on will not reflect the current geometry configuration (plugins etc).

This changeset moved zsh-async worker creation to every render cycle.

### DIFF
--- a/lib/async.zsh
+++ b/lib/async.zsh
@@ -5,6 +5,9 @@ source $GEOMETRY_ROOT/lib/zsh-async/async.zsh 2> /dev/null
     local job="$1" code="$2" output="$3" exec_time="$4" stderr="$5"
     RPROMPT="${(j/::/)output}"
     zle && zle reset-prompt
+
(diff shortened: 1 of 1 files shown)


frm:
I have a question: if we were using a single worker before and submitting jobs, won't this keep several workers alive?

desyncr:
It seems it [reuses](https://github.com/mafredri/zsh-async/blob/master/async.zsh#L388) workers instances by name. Anyways it's better to be on the safer side.

frm (approved):

frm:
I approved but if it re-uses instead of respawning, even better. Although if the context changes (some env variable is redefined) it won't update until it's relaunched.

PR #166, opened 2017-05-31
Add Conda plugin

hanslovsky:
I added a plugin that shows the current `conda` environment if active. This is based on the `virtualenv` plugin.


desyncr:
Thanks @hanslovsky It seems it's OK although I haven't tried it yet. Not quite sure about including it as a built-in plugin, most likely as a custom plugin. What do you think @fribmendes @jedahan

plugins/conda/plugin.zsh
```diff
@@ -0,0 +1,17 @@
+# Color definitions
+DEFAULT_COLOR=${GEOMETRY_COLOR_PROMPT:-green}
```
jedahan: Pretty neat way to inherit from the the environment :)

jedahan:
Thanks for the contribution @hanslovsky! I am OK with us bringing this in as a core plugin, and here is my reasoning:

1. The cost of distributing is low (2 more files on disk)
2. The cost of maintaining is low (this is a simple plugin, and easier to keep in sync with breaking api changes in the future)
3. The benefit for discoverability is high
4. The more we are open to quality contributions, the more people will use geometry :)

So I'd be cool with merging it as-is, and just keeping it out of the default set of enabled plugins.

jedahan:
Generally, this is not the same philosophy I take when evaluating programming languages or frameworks (heck, I used to be an exherbo user...). For something as specific as 'right shell prompt plugins for a particular prompt' I think we should err on just asking our users to edit an environment array over having to discover and clone new repositories.

plugins/conda/plugin.zsh
hanslovsky: Thanks :)

hanslovsky:
Thanks for considering this for a merge @desyncr @jedahan ! In my eyes, the strongest point for adding it as a core plugin is (3) discoverability. As a user, I only use the plugins that are shipped with geometry. That makes it easier to maintain my dotfiles and I am also not aware of any geometry plugins that are not core plugins.

As for (4): Thank you! I have used geometry longer than any other zsh theme so far. Usually I switch frequently.

desyncr:
@hanslovsky We've a point right there about how we publish custom plugins (currently on https://github.com/fribmendes/geometry/wiki/plugins).

hanslovsky:
Thanks for the link!

Update: I am using zplug to manage my zsh config. Do you know how I can add custom geometry plugins through zplug?

desyncr:
@hanslovsky You need to install it as any other plugin, just _after_ geometry. So:

    zplug "fribmendes/geometry"
    zplug "other/geometry-custom-plugin"

Example: https://github.com/desyncr/zshrc/blob/master/.antigenrc#L46 (but with Antigen)

The recommended way for custom plugins is to self-register. So, with only loading them they'll be available to geometry.

hanslovsky:
I had to remove my `~/.zplug` directory and clone it again. Now `desyncr/geometry-pretty-git` works for me.
zplug fails to load `jedahan/geometry-hydrate`:
```bash
[zplug] These repos have failed to load:
- jedahan/geometry-hydrate
```
I will figure out what it is.

Update: It was just a matter of clearing the zplug cache (`zplug clear`). I was able to load geometry-hydrate and this problem was resolved as well.

Now that I know that it's that easy to load a custom plugins, I am also fine with contributing a custom plugin, whichever you prefer.

desyncr (approved):

desyncr:
Updated plugin wiki page with install instructions: https://github.com/fribmendes/geometry/wiki/Plugins#installing

desyncr:
Another approach to this particular plugin is to merge both virtualenv and conda (if they exclude each other) in a unique plugin that handles both instances.

Update: Same as we did with node plugin which uses npm or yarn depending on the context: https://github.com/fribmendes/geometry/tree/master/plugins/node

hanslovsky:
I agree that merging those two would be reasonable. In fact, I first tried to use the `virtualenv` plugin because I assumed that it would cover `conda` as well. It is safe to assume that there is no practical scenario in which both conda and virtualenv are active at the same time.

frm:
Hey guys,

First of all, thanks for the contribution @hanslovsky :heart: 

I see @desyncr's reason for not merging. However, I also tend to agree with @jedahan because of points (1) and (2). To me this is a symptom of something we should address which is: what criteria should we use to decide this? We should discuss that on gitter sometime in the next hangout.

There's also a very important point in:

> As a user, I only use the plugins that are shipped with geometry. That makes it easier to maintain my dotfiles and I am also not aware of any geometry plugins that are not core plugins.

I'm ok with merging it and not making it a default plugin.

However, if it's possible to merge this with the `virtualenv` plugin, that would be the best! I'm not familiar with `conda` so if you confirm that it's highly unlikely that a user has both active simultaneously, I'm up for merging them.

Would like to take this on @hanslovsky ?

hanslovsky:
@fribmendes I don't see any reason why `conda` and `virtualenv` would be used at the same time but I will do some research to make sure there is no such use case.
https://conda.io/docs/_downloads/conda-pip-virtualenv-translator.html
If it happens to be used in conjunction, we could still merge it like this `(<virtualenv-name>:<conda-env-name>)`.

Either way, I will take this on and make changes accordingly.

frm:
Great! Thanks!

desyncr:
@hanslovsky Are you able to work on this? Need any help or guidance? Let me know! 😄

hanslovsky:
@desyncr Have been kind of busy but I have not forgotten this! I should have an update soon.

desyncr:
@hanslovsky Awesome! Thanks for the update! Let us know if you got stuck with anything! :)

hanslovsky:
I got it all worked out. The `virtualenv` plugin now supports both `virtualenv` and `conda`, with individual colors for each of the two. For the case that both are active at the same time (I don't see any reason for that but I wanted to be on the safe side), both will be separated by `${GEOMETRY_VIRTUALENV_CONDA_SEPARATOR}` which defaults to `:`.

I will update the `README.md` and probably update this PR sometime later today.

As a teaser, I made a screenshot:
conda-virtualenv-smaller

desyncr:
Awesome! Amazing work! @hanslovsky Can't wait to get this merged!

hanslovsky:
I updated the PR with 0a28e1b

plugins/virtualenv/plugin.zsh
```diff
@@ -1,15 +1,34 @@
...
 geometry_prompt_virtualenv_render() {
-  local ref=$(basename $VIRTUAL_ENV)
-  echo "$(prompt_geometry_colorize $GEOMETRY_COLOR_VIRTUALENV "(${ref})")"
+
+    if [ -n "${VIRTUAL_ENV}" ]; then
+
+        virtualenv_ref=$(basename $VIRTUAL_ENV)
+        virtualenv_str="$(prompt_geometry_colorize $GEOMETRY_COLOR_VIRTUALENV ${virtualenv_ref})"
+
+        if [ -n "${CONDA_PREFIX}" ]; then
```
desyncr: It looks great and works perfectly. I'm not too fan of duplicating the `CONDA` block though. I'd create 3 separate block for:

1- Generate virtualenv prompt (`if VIRTUAL_ENV`)
2- Generate separator if there's conda AND virtualenv are available (`if VIRTUAL_ENV && CONDA_PREFIX`)
3- Generate conda prompt (`if CONDA_PREFIX`)

Anyways it's a nitpick.

frm (changes requested):
Looks great! But following up on @desyncr's review, do we need the separate `conda` files? You seem to have them all set on `virtualenv`. Or do you intend to allow users to add both `conda` and `virtualenv` (interchangeably) to their plugin list?

desyncr:
I added a comment on the PR. Nothing big or blocking, I'd merge it right away. Only thing missing would be to remove conda plugin (leave only virtualenv that has both env. managers)

hanslovsky:
@fribmendes You are absolutely right, I forgot to remove the `conda`, which is fixed by 78b6a93

frm (commented):
Excellent work. Thanks again.

👍 from my end.

@jedahan do you want to pitch in?

frm (approved):

plugins/virtualenv/plugin.zsh
hanslovsky: Something like this?
```bash
if [ -n "${VIRTUAL_ENV}" ]; then
    virtualenv_ref=$(basename $VIRTUAL_ENV)
    environment_str="$(prompt_geometry_colorize $GEOMETRY_COLOR_VIRTUALENV ${virtualenv_ref})"
fi

if [ -n "${VIRTUAL_ENV}" -a -n "${CONDA_PREFIX}" ]; then
    environment_str="${environment_str}${GEOMETRY_VIRTUALENV_CONDA_SEPARATOR}"
fi

if [ -n "${CONDA_PREFIX}" ]; then
    conda_ref="$(basename $CONDA_PREFIX)"
    environment_str="${environment_str}$(prompt_geometry_colorize $GEOMETRY_COLOR_CONDA ${conda_ref})"
fi
echo "${environment_str}"
```

Let me know if you'd like me to change it before merging.

desyncr: Yeah, exactly what I meant :)

desyncr:
@hanslovsky Tried it and it works flawlessly. Good work!

plugins/virtualenv/plugin.zsh
desyncr: Yes, please go ahead.

hanslovsky: I updated `plugins.zsh` in 4fa1422

desyncr:
I'm gonna go ahead and merge this. We can take a look at minor issues in future iterations if necessary.

Thanks @hanslovsky for the amazing work!

PR #170, opened 2017-06-21
Fix env. var. name for multiline prompts in doc

redianthus:
Hi,

The env. var. name given in the doc for the multiline prompt was false, it's `GEOMETRY_SYMBOL_RPROMPT`, not `GEOMETRY_SYMBOL_PROMPT2`.


desyncr:
Good catch! Thanks!

PR #171, opened 2017-06-21
Replace source by .

redianthus:
It doesn't make a difference with zsh, but `.` is POSIX and `source` is not. So it may be better to tell people to use `.` as they'll learn something that is portable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Clone this repository as follows:
 
 Then add it to your `.zshrc` configuration:
 
-    source /path/to/geometry/geometry.zsh
+    . /path/to/geometry/geometry.zsh
 
 ## Dependencies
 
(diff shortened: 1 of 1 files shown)


frm:
Would this make in difference since `source` is used throughout the project?

desyncr:
We'd have to update the wiki pages as well.

redianthus:
I can do it if you agree that it's a good change.

desyncr:
I have no strong opinions about it. If nobody is against it let go for it. @fribmendes @jedahan

frm:
I don't oppose, as long as we define a convention and stick to it 😛

jedahan:
If geometry's goal was to be portable across non-zsh shells, I would be for it, but there are many other places that use zsh-specific stuff that would make that goal tricky. In addition, `source` is easier to read, and fix if misspelled than `.` for newcomers. Though I am glad that people are considering portability, I'm a weak 'no' on this change for the previous reasons.

desyncr:
I do agree with @jedahan.

@zapashcanon Thanks for contributing anyways!

PR #172, opened 2017-06-21
Add a note about setting var. when using antigen

redianthus:
Hi,

Something like:

```sh
. /usr/share/zsh-antigen/antigen.zsh

GEOMETRY_SYMBOL_PROMPT="λ"

antigen use oh-my-zsh
antigen theme frmendes/geometry
```

won't work. But:

```sh
GEOMETRY_SYMBOL_PROMPT="λ"

. /usr/share/zsh-antigen/antigen.zsh

antigen use oh-my-zsh
antigen theme frmendes/geometry
```

is ok. So, I just added a note about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ geometry was built with easy configuration in mind. The best way to do so is by
 [using environment variables](https://github.com/frmendes/dotfiles/blob/7f448626e1c6e9c0ab7b474c5ff2c1939b64b7d2/system/prompt.zsh#L18-L24).
 
 Pretty much everything in geometry can be changed by setting a variable **before
-you load the theme**.
+you load the theme**. Note that if you use antigen, you must set them before sourcing `antigen.zsh`.
 
 The default options try to balance the theme in order to be both lightweight and contain useful features.
 


frm:
Hi!

Thanks for this. Before merging, are you sure this just applies to antigen? Wouldn't it be similar in `zplug` (iirc)?

redianthus:
I don't use/know zplug, so can't help here, sorry !

desyncr:
It may be too specific to the plugin manager but I don't see too much of a trouble with adding it. 👍

frm:
Oh, yeah, I'm up for merging! I was just wondering if we should change it to be broader, without even specifying a package manager. As an example: "(...) you should set them before loading the theme in your package manager (e.g. antigen)"

desyncr:
Agree on that point.

redianthus:
The fact is that it's not just about loading the theme in the package manager. In my first example, I kinda set them before loading the theme with the package manager but it doesn't work.

desyncr:
@zapashcanon Can you clarify what works and what doesn't work?

I'm up for merging but text needs to be reworked in a generic way (example above is OK).

redianthus:
It's very clear in my first post, no ?

You must set env. var. before loading antigen. Loading them before `antigen theme frmendes/geometry` is not enough.

desyncr:
@zapashcanon What wasn't clear was your previous comment. Otherwise I wouldn't ask.

redianthus:
Hum ok. It was just another way to explain what's in the first post.

frm:
Any updates on this?

desyncr (changes requested):
Re word in a non plugin specific way: "(...) you should set them before loading the theme in your package manager (e.g. antigen)"

redianthus:
It's not really about setting them before loading the theme. (This is already in the `README`).

It's about setting them before "loading" the package manager. Sourcing `antigen.zsh` for antigen e.g., but I don't know how other package manager are loaded/work, so it may not be the case.

frm (approved):
We can merge this, I think. It's really a matter of semantics and if the issue arises in other package managers, we can update it.

Sorry for taking so long to reply @zapashcanon. Tough couple of months.

desyncr:
@zapashcanon This seems to be an issue with this particular package manager rather than with geometry itself.

We think geometry should be agnostic to a package manager. It adds various install methods in its `README.md` but just as a convenience.

frm:
@desyncr yeah, but if it affects users we might as well add in that information. It's not a big deal and it can avoid issues in the future for anyone who tries to install it.

Although I just noticed it's kinda lost in the middle of the `README`. I'd rephrase it and put it in the `antigen` specific section. What do you think?

frm:
@zapashcanon can you take a look at my last comment?

desyncr:
@fribmendes @zapashcanon The issue with Antigen is resolved so this gotcha is not necessary anymore.

frm:
Should we close this then? @desyncr

desyncr:
Yes. Thanks @zapashcanon for the work. It's currently solved in Antigen itself.

PR #177, opened 2017-08-09
Add file/conflict indicators

frm:
I used to get confused a lot regarding which number was the file count and which number was the total conflict count.

I decided to add two indicators `f` for `files` and `c` for `conflicts`.

Preview here:

screen shot 2017-08-09 at 17 50 56

I'm was thinking about extracting the `f/c` indicator into variables but it's something so small and nitpicky that I'm not sure I should do it. Thoughts on that?


jedahan (approved):
I'm fine with the PR as is. No need for f/c to be configurable until someone asks for it

jedahan:
Plus its a good first time contributor change

PR #178, opened 2017-08-26
Adding contribute.md file

shridharmanvi:
Hey guys, I have added a contribute file. Let me know if it needs any changes.

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -14,7 +14,7 @@ When contributing to this repository, please first discuss the change you wish t
 
 ## New plugin template
 
-Details about creating a new plugin are listed [here](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md). See below for a common template that can be followed while creating a new plugin.
+Details about creating a new plugin are listed [here](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md) and a wiki using third party plugins are [here](https://github.com/geometry-zsh/geometry/wiki/plugins) . See below for a common template that can be followed while creating a new plugin.
 
 * To create a new plugin, you will need a setup, check, and a render function, with the plugin name on them
 * Your new plugin needs to be a plugin.zsh file placed inside a directory with the name of the plugin
@@ -22,7 +22,7 @@ Details about creating a new plugin are listed [here](https://github.com/geometr
 * Add your plugin directory to `geometry/plugins/` folder
 
 
-```
+``` zsh
 # Step1: Define your variables above all 
 GEOMETRY_<PLUGIN_NAME>_VAR1=${GEOMETRY_<PLUGIN_NAME>_VAR1:-"VALUE1"}
 GEOMETRY_<PLUGIN_NAME>_VAR2=${GEOMETRY_<PLUGIN_NAME>_VAR2:-"VALUE2"}
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -1,0 +1,49 @@
+# Contributing
+
+When contributing to this repository, please first discuss the change you wish to make/new feature you want to add via issue, or any other method with the owners of this repository. 
+
+
+## Pull Request Process
+
+1. Ensure any install or build dependencies are removed before the end of the layer when doing a 
+   build.
+2. Update the CHANGELOG.md with details of changes to the interface, this includes new environment 
+   variables, exposed ports, useful file locations and container parameters.
+3. Increase the version numbers in any examples files and the README.md to the new version that this
+   Pull Request would represent.
+4. Your PR will be merged (if accepted) once it is reviewed by other developers.
+5. It is advised to create a PR if you are stuck (so others can help you) or when your feature is complete and tested. 
+
+## New plugin template
+
+Details about creating a new plugin are listed [here](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md). See below for a common template that can be followed while creating a new plugin.
+
+* Add your plugin to `geometry/plugins/` folder 
+* To create a new plugin, you will need a setup, check, and a render function, with the plugin name on them.
+
+```bash
+# Step1: Define your variables above all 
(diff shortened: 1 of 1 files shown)


CONTRIBUTE.md
```diff
@@ -0,0 +1,49 @@
+# Contributing
+
+When contributing to this repository, please first discuss the change you wish to make/new feature you want to add via issue, or any other method with the owners of this repository. 
+
+
+## Pull Request Process
+
+1. Ensure any install or build dependencies are removed before the end of the layer when doing a
```
desyncr: Step 1 and 2 seems to focus on Docker rather than geometry.

desyncr (changes requested):
Adding a link to the [plugin docs](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md) would be a good addition.

shridharmanvi:
I have used this link as reference https://gist.github.com/PurpleBooth/b24679402957c63ec426

_Sent from my OnePlus ONEPLUS A5000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

CONTRIBUTE.md
```diff
@@ -0,0 +1,49 @@
...
+5. It is advised to create a PR if you are stuck (so others can help you) or when your feature is complete and tested. 
+
+## New plugin template
+
+Details about creating a new plugin are listed [here](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md). See below for a common template that can be followed while creating a new plugin.
+
+* Add your plugin to `geometry/plugins/` folder 
+* To create a new plugin, you will need a setup, check, and a render function, with the plugin name on them.
+
+```bash
```
frm: `bash` 😞

CONTRIBUTE.md
```diff
@@ -0,0 +1,49 @@
+# Contributing
+
+When contributing to this repository, please first discuss the change you wish to make/new feature you want to add via issue, or any other method with the owners of this repository.
```
frm, line 3: I know you got this from the template but I'm not sure we should include this. A lot of good PRs we have received have not been previously discussed and I think enforcing this is to not promote creativity. I'd rather not enforce it.

Any thoughts? @desyncr @jedahan

CONTRIBUTE.md
```diff
@@ -0,0 +1,49 @@
...
+   Pull Request would represent.
+4. Your PR will be merged (if accepted) once it is reviewed by other developers.
+5. It is advised to create a PR if you are stuck (so others can help you) or when your feature is complete and tested. 
+
+## New plugin template
+
+Details about creating a new plugin are listed [here](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md). See below for a common template that can be followed while creating a new plugin.
+
+* Add your plugin to `geometry/plugins/` folder 
+* To create a new plugin, you will need a setup, check, and a render function, with the plugin name on them.
```
frm: There is also a structure that users should follow. The plugin must be in a `plugin.zsh` file inside a directory with the name of the plugin.

Users are also advised to include a README with the plugin behavior as well as instructions to use and any customization options.

CONTRIBUTE.md
```diff
@@ -0,0 +1,49 @@
...
+When contributing to this repository, please first discuss the change you wish to make/new feature you want to add via issue, or any other method with the owners of this repository. 
+
+
+## Pull Request Process
+
+1. Ensure any install or build dependencies are removed before the end of the layer when doing a 
+   build.
+2. Update the CHANGELOG.md with details of changes to the interface, this includes new environment 
+   variables, exposed ports, useful file locations and container parameters.
+3. Increase the version numbers in any examples files and the README.md to the new version that this
```
frm: We aren't exactly doing semver, so this would not be required to happen. Thoughts? @desyncr @jedahan

frm (changes requested):
This is a great start!

Besides what @desyncr said, I would also include references to the wiki regarding third-party plugins. You can find them [here](https://github.com/geometry-zsh/geometry/wiki/plugins)

shridharmanvi:
@desyncr There is already a link to plugin docs on line 19.

CONTRIBUTE.md
```diff
@@ -5,23 +5,24 @@ When contributing to this repository, please first discuss the change you wish t
...
 
-* Add your plugin to `geometry/plugins/` folder 
-* To create a new plugin, you will need a setup, check, and a render function, with the plugin name on them.
+* To create a new plugin, you will need a setup, check, and a render function, with the plugin name on them
+* Your new plugin needs to be a plugin.zsh file placed inside a directory with the name of the plugin
+* Users are advised to also place a README.md file inside the plugin directory with details about its behaviour, use and customization options  
+* Add your plugin directory to `geometry/plugins/` folder
+
 
-```bash
```
frm: Oh, I was fine with bash for syntax highlight. I'd actually prefer to have it. I was just commenting because this is a zsh theme. You can also put `zsh` instead of `bash`.

Also, fun fact: in GitHub syntax highlighting, `zsh` and `bash` are both aliases of `shell`.

frm (approved):

CONTRIBUTE.md
shridharmanvi: Sure, should I go ahead and take it off? Do you want to replace it with something else?

CONTRIBUTE.md
shridharmanvi: Gotcha. Thanks @fribmendes

CONTRIBUTE.md
frm: I'd prefer taking it off if that's ok with @jedahan and @desyncr

shridharmanvi: I have added a link to the plugin wiki as you called out earlier.

jedahan (approved):
This is a fine first version of contribute.md

shridharmanvi:
Hey @desyncr let me know if you need any other changes.

desyncr (approved):

desyncr:
@shridharmanvi Looking great! Sorry for the long delay in my review! :) Hope to see more of yours contributions!

PR #184, opened 2017-11-18
WIP: Add vi-mode as built-in plugin

### LINKED ISSUE #165: Plugin Request: Vi-Mode
Zshell has the ability to go into vi-mode with `bindkey -v` settings. If you are using `oh-my-zsh`, this feature can be enabled by including the [vi-mode](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/vi-mode/vi-mode.plugin.zsh) plugin.

However, the interest with `geometry` is to port the visual indicator over.  [vi-mode](https://github.com/robbyrussell/oh-my-zsh/blob/master/plugins/vi-mode/vi-mode.plugin.zsh#L43) implements its own visual indicator function `vi_mode_prompt_info()` where it toggles between `vimcmd` and `main` mode indicating `NORMAL` and `INSERT` mode respectively.

The request here is to include the visual indicator as a plugin in `geometry`. Thank you in advance.

\cc @fribmendes, @desyncr

desyncr:
- [ ] Add a proper README.md
  - [ ] Fix issue with vi-mode not showing on start up / after command
  - [ ] Use icon/glyp rather than text

screenshot from 2017-11-17 23-17-18

Fixes #165

### DIFF
--- a/plugins/vi-mode/plugin.zsh
+++ b/plugins/vi-mode/plugin.zsh
@@ -1,10 +1,18 @@
-zle -N zle-keymap-select geometry_prompt_vi-mode_render
 
+# For some reason this line doesn't work inside a function, otherwise
+# it should be located in the _setup function.
+zle -N zle-keymap-select geometry_prompt_vi-mode_render
 geometry_prompt_vi-mode_setup() {}
+
 geometry_prompt_vi-mode_check() {}
+
 geometry_prompt_vi-mode_render() {
-    RPROMPT=${${RPROMPT/\[INSERT\] /}/\[NORMAL\] /}
-    local KEY=${${KEYMAP/main/INSERT}/vicmd/NORMAL}
-    RPROMPT="[$KEY] $RPROMPT"
-    zle && zle reset-prompt || echo -n $RPROMPT
+  # Removes previously state
+  RPROMPT=${${RPROMPT/\[INSERT\] /}/\[NORMAL\] /}
+  # swaps main/INSERT, vicmd/NORMAL
+  local KEY=${${KEYMAP/main/INSERT}/vicmd/NORMAL}
+  # Formats RPROMPT with mode + rest of prompt
+  RPROMPT="[$KEY] $RPROMPT"
+  # Resets prompt
+  zle && zle reset-prompt
 }
--- a/plugins/vi-mode/plugin.zsh
+++ b/plugins/vi-mode/plugin.zsh
@@ -1,0 +1,10 @@
+zle -N zle-keymap-select geometry_prompt_vi-mode_render
+
+geometry_prompt_vi-mode_setup() {}
+geometry_prompt_vi-mode_check() {}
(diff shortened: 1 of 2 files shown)


plugins/vi-mode/plugin.zsh
```diff
@@ -0,0 +1,18 @@
...
+
+geometry_prompt_vi-mode_render() {
+  # Removes previously state
+  RPROMPT=${${RPROMPT/\[INSERT\] /}/\[NORMAL\] /}
+  # swaps main/INSERT, vicmd/NORMAL
+  local KEY=${${KEYMAP/main/INSERT}/vicmd/NORMAL}
+  # Formats RPROMPT with mode + rest of prompt
+  RPROMPT="[$KEY] $RPROMPT"
+  # Resets prompt
+  zle && zle reset-prompt
```
frm, line 17: Won't this do an extra redraw per prompt line?

plugins/vi-mode/plugin.zsh
```diff
@@ -0,0 +1,18 @@
+
```
frm, line 1: Extra newline

frm (changes requested):
Looking good but should we clear that checklist first? :)

plugins/vi-mode/plugin.zsh
desyncr: Yes. I think it can be avoided though.

desyncr:
Yes indeed, this is a WIP. Also any idea about icon?

frm:
I missed the icon question before. It's not really easy to answer since there is no clear character that can be associated with vim. One alternative would be to change the prompt symbol (color or actual character) but I'm not sure that's possible or even desirable since it means that it has to be re-rendered).

plugins/vi-mode/plugin.zsh
jedahan: How can it be avoided? can we just set a dirty bit somewhere?

DarwinSenior:
Hi,
I was wondering if I could take this task?
I think the changing of the prompt symbol is desirable as it is consistent with the minimalistic approach of this theme.

I happened to solve this with the following configuration.

```zsh
setopt PRUMPT_SUBST
GEOMETRY_SYMBOL_ROOT='${${KEYMAP/vicmd/▼}/(main|viins)/▲}'
GEOMETRY_SYMBOL_PROMPT='${${KEYMAP/vicmd/▼}/(main|viins)/▲}'
GEOMETRY_SYMBOL_EXIT_VALUE='${${KEYMAP/vicmd/▽}/(main|viins)/△}'
function zle-line-init zle-keymap-select {
  zle reset-prompt
}
zle -N zle-line-init
zle -N zle-keymap-select
```
And would it be better to add the following to the path plugin rather than to create a new plugin?

jedahan:
Only problem I have with doing this in the PATH plugin is in the case where someone would want to disable this functionality. Otherwise I like your solution and ideas.

PR #194, opened 2017-12-11
Make sure to sanitize pinned plugins in plugin_check (Fixes #192)

jedahan:
I had to do this to fix `+rustup` in my config

### DIFF
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -72,7 +72,7 @@ geometry_plugin_register() {
     return 1
   fi
 
-  if geometry_prompt_${plugin#+}_setup $ctx; then
+  if $plugin_setup_function $ctx; then
     # Register plugin in $ctx with '+' for pinning
     _ctx_plugins+=$plugin
     _GEOMETRY_PROMPT_PLUGINS[$ctx]=${(j/ /)_ctx_plugins}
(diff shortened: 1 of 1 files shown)


lib/plugin.zsh
```diff
@@ -72,7 +72,7 @@ geometry_plugin_register() {
     return 1
   fi
 
-  if geometry_prompt_${plugin#+}_setup $ctx; then
+  if $plugin_setup_function $ctx; then
```
desyncr, line 75: Isn't this sanitized on [this](https://github.com/geometry-zsh/geometry/pull/194/files#diff-fa15ebb02da94453c62e6471145db51aR69) line?

jedahan: this line isn't the fix, this was just to reduce redundancy since we have access to the local anyway

desyncr: oh I see now! :+1:

desyncr (approved):

frm (approved):

desyncr:
I tried it and I see no breaking changes. What's a test case?

desyncr:
btw, my configs:

```
GEOMETRY_PROMPT_PLUGINS_SECONDARY=(exec_time jobs +node git +hostname)
GEOMETRY_PROMPT_PLUGINS_PRIMARY=(path)
```

jedahan:
On a fresh install of geometry, minimal zshrc with `GEOMETRY_PROMPT_PLUGINS_SECONDARY=(+rustup)` before I would get errors 'rustup plugin not found'. Now I don't.

desyncr:
Mergin then!

Issue #197, opened 2018-01-09
Change custom plugins examples and guidelines


desyncr:
Current documentation suggests for plugins to [self register](https://github.com/geometry-zsh/geometry/blob/master/plugins/README.md#full-working-example). Back then it was a good idea as it loaded plugins configuration-free.

Current implementation (right and left prompt pluggable) enables further customization and it makes sense to be able to tell the plugins to either load at the right or left prompt:

    GEOMETRY_PROMPT_PLUGINS_SECONDARY=(custom_plugin ...)
    # Or
    GEOMETRY_PROMPT_PLUGINS_PRIMARY=(custom_plugin ...)

This conflicts with the current plugin loading approach as by default self-registration uses the `SECONDARY` prompt as target.

We do support both registration methods but should update the documentation to use the configuration approach rather than self-registration.

An example configuration would be such as:

    # Add the plugin to a prompt, in this example "secondary"
    GEOMETRY_PROMPT_PLUGINS_SECONDARY=(custom_plugin ...)

    # Source the plugin script BEFORE geometry
    source custom_plugin.zsh

    # Finally source geometry theme
    source geometry.zsh

desyncr:
Actually there's a little fix needed for custom-plugins to be able to be configured by context.

frm:
@desyncr can we label this as a nice first PR for people?

jedahan:
Is there any case that someone would want to source a geometry plugin and it not be present in secondary prompt?

I would rather have us auto-put in secondary prompt, *UNLESS* it is defined to be in primary prompt by user.

So default configureless, but if you configure it overrides.

frm:
Is there a way for this to happen though? You have to explicitly set it to the secondary or primary variable, right?

jedahan:
we check in register function if plugin is currently in PRIMARY. if it is, we do nothing. if its not, we register to SECONDARY

i havent looked at the code in a while though, so i might be missing something obvious

PR #198, opened 2018-01-09
Custom plugin context-awareness

desyncr:
See https://github.com/geometry-zsh/geometry/issues/197


jedahan:
This change means if you try registering a plugin without sourcing it, we will error? I'm cool with that.

jedahan (approved):
better to explicitly fail 'hey you tried to register a plugin we dont know about!'

frm (approved):

jedahan:
This breaks some configs so lets not release until its fixed. #206 should fix it.

PR #201, opened 2018-01-31
add git stash indication

ev-agelos:
The icon doesn't look nice with the rest I just chose one :/
And I am not sure for the color too :D


plugins/git/plugin.zsh
```diff
@@ -172,13 +176,17 @@ geometry_prompt_git_render() {
     fi
   fi
 
+  if git rev-parse --quiet --verify refs/stash >/dev/null; then
```
jedahan: I would check `PROMPT_GEOMETRY_GIT_SHOW_STASHES` here first

plugins/git/plugin.zsh
```diff
@@ -29,6 +32,7 @@ PROMPT_GEOMETRY_GIT_CONFLICTS=${PROMPT_GEOMETRY_GIT_CONFLICTS:-false}
 PROMPT_GEOMETRY_GIT_TIME=${PROMPT_GEOMETRY_GIT_TIME:-true}
 PROMPT_GEOMETRY_GIT_TIME_LONG_FORMAT=${PROMPT_GEOMETRY_GIT_TIME_LONG_FORMAT:-false}
 PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY=${PROMPT_GEOMETRY_GIT_TIME_SHOW_EMPTY:-true}
+PROMPT_GEOMETRY_GIT_STASHES=${PROMPT_GEOMETRY_GIT_STASHES:-false}
```
jedahan: can we name this `PROMPT_GEOMETRY_GIT_SHOW_STASHES` and set the default to `true` ?

plugins/git/plugin.zsh
```diff
@@ -172,13 +176,17 @@ geometry_prompt_git_render() {
...
+  fi
+
   local render="$(prompt_geometry_git_symbol)"
 
   if [[ -n $render ]]; then
     render+=" "
   fi
 
-  render+="$(prompt_geometry_git_branch) ${conflicts}${GEOMETRY_GIT_SEPARATOR}${time} $(prompt_geometry_git_status)"
+  render+="$(prompt_geometry_git_branch) ${conflicts}${GEOMETRY_GIT_SEPARATOR}${time} $(prompt_geometry_git_status) ${stashes}"
```
jedahan: I'd like to see what this looks like before the status, with the other 'optional' parts.
Also we should include the space / GEOMETRY_GIT_SEPARATOR as part of the `stashes=` assignment so if stashes is unassigned we don't have a trailing space.

jedahan (approved):
Great addition! Do you have the time to update the PR with my suggestions or should I?

jedahan:
Trying to find a replacement for the pin icon that keeps simple, solid geometry - these seem nice:

# `●` `✦` `╸` `⚈` 

I want to emphasize that stashes are not as big a deal as conflicts too

plugins/git/plugin.zsh
ev-agelos: Sure!

plugins/git/plugin.zsh
ev-agelos: Right. Ill update it

ev-agelos:
I will update the PR with your suggestions, thanks for the review!
It is my 2nd contribution and I would like to to do it from start to finish :D

jedahan:
testing today

jedahan:
Thanks for your patience and the contribution!

ev-agelos:
Awesome! Thanks a lot too for you guidance!

Issue #196, opened 2018-01-09
Change node plugin default color


desyncr:
Plugin symbol's color currently is [black](https://github.com/geometry-zsh/geometry/blob/master/plugins/node/plugin.zsh#L2), which is kind of hard to see. I suggest we change the default to green instead.

Current (black):

![screen shot 2018-01-09 at 3 55 55 pm](https://user-images.githubusercontent.com/1857707/34737705-a6e591c6-f555-11e7-9ca4-e8c041f4d214.png)

Suggested (green):

![screen shot 2018-01-09 at 3 57 48 pm](https://user-images.githubusercontent.com/1857707/34737782-de93ce44-f555-11e7-9468-e48b8e26e642.png)

frm:
@desyncr can we label this as an easy first PR?

PR #205, opened 2018-03-13
Change node plugin default color

mustaqimM:
Solves #196


jedahan:
lgtm, thanks for the help @mustaqimM !

PR #206, opened 2018-03-17
Allow custom plugins that autoregister to be in PRIMARY

### LINKED ISSUE #197: Change custom plugins examples and guidelines

jedahan:
Right now, custom plugins that autoregister go in the secondary
(right hand) prompt. With this change, if someone puts the
custom plugin in GEOMETRY_PROMPT_PLUGINS_PRIMARY, we will
respect that decision.

This should fix #197 and revert #198

### DIFF
--- a/lib/plugin.zsh
+++ b/lib/plugin.zsh
@@ -40,12 +40,9 @@ geometry_plugin_setup() {
   for ctx in $GEOMETRY_PROMPT_CTX; do
     _ctx_plugins=(${(s/ /)GEOMETRY_PROMPT_PLUGINS[$ctx]})
     for plugin in $_ctx_plugins; do
-      # Source built-in plugin if necessary, custom plugins should be already
-      # sourced by the user, otherwise `geometry_plugin_register` will raise and error
-      test -f "$GEOMETRY_ROOT/plugins/${plugin#+}/plugin.zsh" && source $_
-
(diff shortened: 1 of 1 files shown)


jedahan:
This means we can continue with autoregister, while still allowing people to set plugins however they want

frm (approved):

Issue #3, opened 2016-05-19
Tab completion problem


terry182:
I am having a problem using this theme when I try to tab complete

so it's like
`
 ▲ .oh-my-zsh/custom/themes ls`
and after tab completion it becomes
`
 ▲ .oh-my-zsh/custom/themesls`

I tried to modify the theme but failed. Really can't figure out where the problem is

frm:
What OS and terminal emulator are you running, so I can try to reproduce the issue?

terry182:
@frmendes  I am using OSX El Capitan 10.11.5 (15F34). The emulator is the terminal.app (?

frm:
@terry182 I can't seem to reproduce it. I have the same OS version and tried on Terminal.app. It works on both Terminal and iTerm2 on my end. Can you try and disable other plugins you have (preferably one at a time) to see if there's some side effect created by the theme?

terry182:
@frmendes I didn't use any plugin as I just clean install my Macbook. But I notice that I am just using it as a theme, will that be the problem?

I just download the `geometry.zsh` and put it in `~/.oh-my-zsh/custom/themes` and rename it. 

FYI, I reproduce the result by pressing `l`, `s`, `Tab` in my end. Maybe this could help

terry182:
I found the problem happens to be the international setting `Unicode East Asian Ambiguous characters are wide` of terminal.

I guess this issue can be closed.

IanCStewart:
@frmendes it seems that tab completion adds a character to the sentence 
![screen shot 2016-07-25 at 11 20 36](https://cloud.githubusercontent.com/assets/14125280/17096496/f517979a-5259-11e6-85fd-55b1af1436b1.png)
if you type anything and then `tab` it changes from for example `git` to `ggit`
_this is just visual, the commands still get send without the extra_

frm:
Does this happen by just typing `<command> <tab>`? What encoding are you using?

IanCStewart:
it happens automatic when u type `<command> <tab>` & when you type `<half command> <tab>` and this half command has lots of possibilities. but not when you type `<half command> <tab>` and there  is only one possibility. my encoding must be on the default don't know where to find this.

frm:
What terminal emulator and OS are you using?

IanCStewart:
iTerm and osx El Capitan

frm:
[...]

PR #216, opened 2018-06-28
Ignore character modifying codes in the prompt.

frm:
Why:

* If a user set `GEOMETRY_SYMBOL_PROMPT="\e[1mX\e[0m"`, it would be
expected that X was bold.
* However, even though X would be bold, zsh would assume a much larger
prompt character size of 11, which would in turn, mess up with the
autocomplete.

This change addresses the need by:

* Performing substitution on the prompt symbol before counting the
characters.

Notes:

* This should fix the issue brought up by @n-a-t-h-i-a-s in #3. See [this comment](https://github.com/geometry-zsh/geometry/issues/3#issuecomment-400657117).
* I'm not completely satisfied with this solution since it doesn't fully address the issue. It's very specific to this use case. However, as more use cases come by with those types of errors, we can refine it.
* It might be worthwhile to look at what other prompts are doing, such as `adam2` in [this Stack Overflow question](https://stackoverflow.com/questions/17232962/zsh-variable-substitution-in-adam-explanation).

### DIFF
--- a/plugins/path/plugin.zsh
+++ b/plugins/path/plugin.zsh
@@ -55,10 +55,12 @@ geometry_prompt_path_setup() {
 geometry_prompt_path_check() {}
 
 geometry_prompt_path_render() {
+  local prompt_symbol=""
+
   if [ $? -eq 0 ] ; then
-    PROMPT_SYMBOL=$GEOMETRY_SYMBOL_PROMPT
+    prompt_symbol=$GEOMETRY_SYMBOL_PROMPT
   else
-    PROMPT_SYMBOL=$GEOMETRY_SYMBOL_EXIT_VALUE
+    prompt_symbol=$GEOMETRY_SYMBOL_EXIT_VALUE
   fi
 
   if $PROMPT_GEOMETRY_COLORIZE_ROOT && [[ $UID == 0 || $EUID == 0 ]]; then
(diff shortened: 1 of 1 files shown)


frm:
One possible alternative would be to recommend users to use other alternatives for bold characters. `%BX%f` would be equivalent and would allow us to use a substitution mechanism similar to the `adam2` one I linked. It feels a bit better do that instead of this, which feels a bit hackish.

It would a matter of changing one line to

```zsh
local symbol_width="${#${(S%%)prompt_symbol//(\%([KF1]|)\{*\}|\%[Bbkf])}}"
```

plugins/path/plugin.zsh
```diff
@@ -70,5 +72,12 @@ geometry_prompt_path_render() {
     dir=$(basename $PWD)
   fi
 
-  echo "$GEOMETRY_PROMPT_PREFIX$GEOMETRY_PROMPT_PREFIX_SPACER%${#PROMPT_SYMBOL}{%(?.$GEOMETRY_PROMPT.$GEOMETRY_EXIT_VALUE)%}$GEOMETRY_SYMBOL_SPACER%F{$GEOMETRY_COLOR_DIR}$dir%f$GEOMETRY_DIR_SPACER$GEOMETRY_PROMPT_SUFFIX"
+  local prompt_prefix="$GEOMETRY_PROMPT_PREFIX$GEOMETRY_PROMPT_PREFIX_SPACER"
+
+  local symbol_width="${#${(S%%)prompt_symbol//\\e\[[0-9]m}}"
```
jedahan: Can you add a comment linking to this issue
And mayb a TODO/FIXME comment about the more generic solution?
`local symbol_width="${#${(S%%)prompt_symbol//(\%([KF1]|)\{*\}|\%[Bbkf])}}"`

jedahan (changes requested):

jedahan:
I think this refactor is generally better than our current state, both in that it helps a user, and is easier to read with the split up variables. We can revisit in the future but I am fine with this hack being merged sooner than later.

plugins/path/plugin.zsh
frm: I'd actually prefer that we encourage the use of `%B` for bold, since it goes better with a more generic solution, like the `${(S%%)prompt_symbol//(\%([KF1]|)\{*\}|\%[Bbkf])}` substitution. Updating the PR to use that instead and linking the issue.

frm:
Here's a preview of the support:

screen shot 2018-07-02 at 21 22 26

When merging I'd rather squash and keep the first commit message, as it provides context for everything, when looking through the git log 😄

jedahan:
Ahh squash & merge makes sense. This looks good to me

jedahan (approved):

Issue #224, opened 2018-09-20
Git dirtiness check not working correctly


dmhacker:
Hello! I thought I might bring this to your attention, since it seems that the dirtiness check for git repositories is not working correctly. It's pretty simple to reproduce:

```
mkdir test && cd test
git init
mkdir dir1
mkdir dir2
touch dir1/placeholder
cd dir2
```

At the end of running these commands, you'll notice that the git repository is being marked as clean inside **dir2** when it should be marked as dirty. However, when you `cd ..` outside of **dir2**, the repository is then correctly marked as being dirty.

The problem seems to be arising from a bug in `prompt_geometry_git_status()`. You are running the commands `git status --porcelain --ignore-submodules HEAD` and `git ls-files --others --modified --exclude-standard` to perform the dirtiness check; however, the `ls-files` command implicity runs from the current directory. What you really want is `git ls-files --others --modified --exclude-standard $(git rev-parse --show-toplevel)` which then runs the command from the top level git directory. 

My solution is to is to run a patched version of the function that just uses `git status --porcelain --ignore-submodules` and removes the `ls-files` command. I added a flag `PROMPT_GEOMETRY_GIT_UNTRACKED_DIRTY` that can disable checks for untracked files for large repositories (by simply running a `git diff`). It's the approach that the pure prompt takes. I'll be submitting a pull request shortly that you guys can review.

Love the prompt otherwise though. Keep up the good work!

jedahan:
Hey, I think I used the first fix you described in my branch https://github.com/jedahan/geometry (which also simplifies a lot how to use geometry). Thanks for the report and PR!

jedahan:
fixed in the latest

PR #225, opened 2018-09-20
Fix git dirtiness check & add option to ignore untracked files

dmhacker:
See #224 for a description of the issue. This pull request takes pure's approach to running status checks. I've updated the plugin README accordingly.


jedahan:
This looks good to me, gonna see if we can merge this in soon, or port it to our new architecture if that gets approved. Just didn't wanna leave you hanging on a contribution, thanks for the PR!

jedahan:
Do you have an example huge repository where this can be slow?
I'd love it if the user didn't have to think about another variable - with the right prompt being async, it shouldn't be that bad for it to be slow either way.

jedahan:
Closing in favor of the `mnml` branch of geometry. If it is still relevant, feel free to open and issue and I will port this PR.

PR #227, opened 2018-10-04
doc: Update README to match implementation

drager:
In the code https://github.com/geometry-zsh/geometry/blob/master/plugins/node/plugin.zsh#L5 it's called `GEOMETRY_SYMBOL_PACKAGER_VERSION` so this PR update README to match the code.


jedahan:
Looks great to me thank you @drager !

PR #257, opened 2019-05-22
Add multiline capability to geometry's prompt

dinocosta:
Sometimes I prefer using a multiline prompt, for example when the git branch name is too big. In this situation, I've found myself messing around with geometry in order to use separate the plugins from the user prompt.

As such, this Pull Request adds a `PROMPT_GEOMETRY_MULTILINE` flag which is set to `false` by default. Changing that flag to `true` makes geometry render the prompt in two lines, the first only for the directory and plugins and the second one only for the user prompt.

I've used `GEOMETRY_SYMBOL_SPACER` as the separator between the directory and the plugins but I don't know if this would be the best option or if it would be better to create a new variable just for this separator so I would appreciate some feedback here.

Also, the [contribution guide](https://github.com/geometry-zsh/geometry/blob/master/CONTRIBUTE.md) mentions updating the version but I couldn't find it anywhere except in the changelog. If I forgot to update something feel free to raise that to my attention.


jedahan:
Thanks for making this PR. Unfortunately, we just released v2 of geometry which has a lot of changes to how things work. Would you mind starting a new PR based on the `mnml` branch? I'd be happy to merge if it was straightforward.

dinocosta:
Hi @jedahan thanks for the reply! 👋 
I'll see if I get the time to try and do that :)

PR #259, opened 2019-05-30
Plugin: Exit Status RPROMPT

alxbl:
I couldn't find a clean and modular way to add the exit status to the RPROMPT so I put this quick module together. I'm submitting a PR in case other people could benefit from it.

Currently it only supports one color but it would be easy to add a success color as well for `+exit_status` setups.

Cheers,
Alex


jedahan:
Mind porting to the new `mnml` branch? It should be even simpler :)

alxbl:
Sure, I'll take a look when I have a moment

PR #261, opened 2019-06-09
Enable use of kube-ps1 instead of kube plugin

olegTarassov:
Enable use of [kube-ps1](https://github.com/jonmosco/kube-ps1) that behaves faster and better then geometry plugin
[kube](https://github.com/geometry-zsh/geometry/tree/master/plugins/kube)

This appends on the RPROMP of the PS1 the k8s context.


jedahan:
Now that geometry v2 is released (`mnml` branch), we dont even need this plugin, just `GEOMETRY_RPROMPT+=(kube_ps1)` and it will work!

PR #263, opened 2019-08-21
feat: Added `GEOMETRY_STATUS_SHOW_CODE` and configuration.

alxbl:
As requested, I've ported this to the `mnml` branch. After looking at the `status` function, I figured that it made a lot of sense to have this as an option of status, so I added a configuration to show exit status.

If you'd prefer this to be a standalone function (to make it possible to have it in RPROMPT or INFO instead, let me know.

### DIFF
--- a/functions/geometry_status.zsh
+++ b/functions/geometry_status.zsh
@@ -7,9 +7,6 @@ geometry_status() {
   : ${GEOMETRY_STATUS_SYMBOL_ERROR:=△}
   : ${GEOMETRY_STATUS_SYMBOL_ROOT:=▼}
   : ${GEOMETRY_STATUS_SYMBOL_ROOT_ERROR:=▽}
-  : ${GEOMETRY_STATUS_SHOW_CODE:=false}
-  : ${GEOMETRY_STATUS_CODE_PREFIX:=' '}
-  : ${GEOMETRY_STATUS_CODE_SUFFIX:=' '}
 
   ( ${GEOMETRY_STATUS_SYMBOL_COLOR_HASH:=false} ) && {
     local colors=({1..9})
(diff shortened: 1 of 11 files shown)


jedahan:
Thanks for porting! 

I think standalone is nicer, but I wonder if we should a) be exporting LAST_STATUS, or b) simply setting `?` to what it should be afterwards? If its not too much trouble to have it be a separate function like `geometry_exitcode` that would be awesome, otherwise I will just merge this.

functions/geometry_status.zsh
```diff
@@ -7,6 +7,9 @@ geometry_status() {
   : ${GEOMETRY_STATUS_SYMBOL_ERROR:=△}
   : ${GEOMETRY_STATUS_SYMBOL_ROOT:=▼}
   : ${GEOMETRY_STATUS_SYMBOL_ROOT_ERROR:=▽}
+  : ${GEOMETRY_STATUS_SHOW_CODE:=false}
+  : ${GEOMETRY_STATUS_CODE_PREFIX:=' '}
+  : ${GEOMETRY_STATUS_CODE_SUFFIX:=' '}
```
jedahan: how about using `GEOMETRY_SEPARATOR` instead of `PREFIX`/`SUFFIX`?

jedahan: how about using `GEOMETRY_SEPARATOR` instead of `PREFIX`/`SUFFIX`?

jedahan: how about using `GEOMETRY_SEPARATOR` instead of `PREFIX`/`SUFFIX`?

alxbl:
I pulled the `GEOMETRY_LAST_STATUS` into the precmd hook, which was necessary for it to be readable while building the RPROMPT. I also made exit code into its own function as you said. This way it can be put on either side, and will use `GEOMETRY_SEPARATOR`.

This should be the most modular approach. I'll update the doc in a few minutes.

Cheers,
Alex

alxbl:
One last thing, If you have an idea for a good symbol for the exit code (I was thinking unicode cross?), I can also add the symbol for it.

alxbl:
I tried to rebase this on the latest mnml but had a lot of trouble, so I ended up merging. Conflicts should be fixed. Feel free to squash merge.

jedahan:
Merged it in, thanks for the contribution!

jedahan:
see https://github.com/geometry-zsh/geometry/commit/0721dbbb5028f9b410df7422a3d37e3ef65dc9e0

PR #264, opened 2019-08-27
Feature/kube improvements

olegTarassov:
1) Test KUBECONFIG global variable instead of file
- location of kube/config can be moved or appended using several files
2) Avoid printing kubernetes cluster version unless needed
- keep terminal line to the minimal
3) Put some nice colors
- control context or namespace color
4) Add Separator for geometry_kube

### DIFF
--- a/functions/geometry_kube.zsh
+++ b/functions/geometry_kube.zsh
@@ -3,15 +3,28 @@
 (( $+commands[kubectl] )) || return
 
 geometry_kube() {
-  ( ${GEOMETRY_KUBE_PIN:=false} ) || [[ -f ~/.kube/config ]] || return
+  ( ${GEOMETRY_KUBE_PIN:=false} ) || [[ ! -z "${KUBECONFIG}" ]] || return
 
-  GEOMETRY_KUBE=$(ansi ${GEOMETRY_KUBE_COLOR:=blue} ${GEOMETRY_KUBE_SYMBOL:="⎈"})
+  local geometry_kube=$(ansi ${GEOMETRY_KUBE_COLOR:=blue} ${GEOMETRY_KUBE_SYMBOL:="⎈"})
+  local geometry_kube_separator=${GEOMETRY_KUBE_SEPARATOR:="|"}
 
-  [[ $(kubectl version --client --short) =~ 'Client Version: ([0-9a-zA-Z.]+)' ]]
-  local version=$match[1]
+  if ( ${GEOMETRY_KUBE_VERSION:=true} ); then
+    [[ $(kubectl version --client --short) =~ 'Client Version: ([0-9a-zA-Z.]+)' ]]
+    geometry_kube+=($match[1])
+  fi
 
-  local context="$(kubectl config current-context 2> /dev/null)"
-  local namespace="$(kubectl config view -o "jsonpath={.contexts[?(@.name==\"$context\")].context.namespace}" 2> /dev/null)"
+  local geometry_context="$(kubectl config current-context 2> /dev/null)"
+  local geometry_namespace="$(kubectl config view -o "jsonpath={.contexts[?(@.name==\"$context\")].context.namespace}" 2> /dev/null)"
 
-  echo -n "$GEOMETRY_KUBE $version ($context:${namespace:=default})"
+  if [[ ! -z "${GEOMETRY_KUBE_CONTEXT_COLOR}" ]]; then
+    geometry_context="$(ansi ${GEOMETRY_KUBE_CONTEXT_COLOR:=white} ${geometry_context})"
+  fi
+
+  if [[ ! -z "${GEOMETRY_KUBE_NAMESPACE_COLOR}" ]]; then
+    geometry_namespace="$(ansi ${GEOMETRY_KUBE_NAMESPACE_COLOR:=white} ${geometry_namespace:=default})"
+  else
+    geometry_namespace="${geometry_namespace:=default}"
+  fi
+
+  echo -n "${geometry_kube_separator}${geometry_kube} ${geometry_context}:${geometry_namespace}${geometry_kube_separator}"
 }
--- a/functions/geometry_kube.zsh
+++ b/functions/geometry_kube.zsh
@@ -5,26 +5,36 @@
 geometry_kube() {
   ( ${GEOMETRY_KUBE_PIN:=false} ) || [[ ! -z "${KUBECONFIG}" ]] || return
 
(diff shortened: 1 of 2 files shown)


functions/geometry_kube.zsh
```diff
@@ -3,15 +3,28 @@
...
-  local version=$match[1]
+  if ( ${GEOMETRY_KUBE_VERSION:=true} ); then
+    [[ $(kubectl version --client --short) =~ 'Client Version: ([0-9a-zA-Z.]+)' ]]
+    geometry_kube+=($match[1])
+  fi
 
-  local context="$(kubectl config current-context 2> /dev/null)"
-  local namespace="$(kubectl config view -o "jsonpath={.contexts[?(@.name==\"$context\")].context.namespace}" 2> /dev/null)"
+  local geometry_context="$(kubectl config current-context 2> /dev/null)"
+  local geometry_namespace="$(kubectl config view -o "jsonpath={.contexts[?(@.name==\"$context\")].context.namespace}" 2> /dev/null)"
```
jedahan: maybe `kube_context` and `kube_namespace`?

jedahan: Though I can see an argument for this being geometry specific, not kube_specific. Either way, these variables should be 'hidden' unless something breaks, in which case letting the person know 'this is not kube broken, its your prompt' is less panic-inducing.

functions/geometry_kube.zsh
```diff
@@ -3,15 +3,28 @@
...
 
-  echo -n "$GEOMETRY_KUBE $version ($context:${namespace:=default})"
+  if [[ ! -z "${GEOMETRY_KUBE_CONTEXT_COLOR}" ]]; then
+    geometry_context="$(ansi ${GEOMETRY_KUBE_CONTEXT_COLOR:=white} ${geometry_context})"
+  fi
+
+  if [[ ! -z "${GEOMETRY_KUBE_NAMESPACE_COLOR}" ]]; then
```
jedahan, line 32: does this make `${GEOMETRY_KUBE_NAMESPACE_COLOR:=white}` redundant?

I think this entire test and else branch can be removed and have the same effect?

options.md
```diff
@@ -11,19 +11,19 @@ and functionality.
...
+  - [`geometry_git_rebase()`](#geometrygitrebase)
+  - [`geometry_git_remote() `](#geometrygitremote)
+  - [`geometry_git_conflicts()`](#geometrygitconflicts)
+  - [`geometry_git()`](#geometrygit)
+  - [`geometry::git_wrapper`](#geometrygitwrapper)
   - [`geometry_hg`](#geometryhg)
   - [`geometry_hostname`](#geometryhostname)
   - [`geometry_jobs`](#geometryjobs)
-  - [`geometry_kube`](#geometrykube)
+  - [`geometry_kube`](#geometry_kube)
```
jedahan, line 26: nice catch!

functions/geometry_kube.zsh
```diff
@@ -3,15 +3,28 @@
...
+    geometry_namespace="$(ansi ${GEOMETRY_KUBE_NAMESPACE_COLOR:=white} ${geometry_namespace:=default})"
+  else
+    geometry_namespace="${geometry_namespace:=default}"
+  fi
+
+  echo -n "${geometry_kube_separator}${geometry_kube} ${geometry_context}:${geometry_namespace}${geometry_kube_separator}"
```
jedahan: So whats neat is that zsh supports joins that are a bit smarter - you can look at `geometry::wrap` for how it works, but if you have an array of things you want to join with a string, you can do something like (untested)
```zsh
geometry_kube_info=($geometry_kube ${geometry_context}:${geometry_namespace})
echo "${(psj.${geometry_kube_separator}.)geometry_kube_info}"
```

jedahan (approved):

functions/geometry_kube.zsh
olegTarassov: You are correct, the default white is redundant.

Secondly, technically we can remove the if else and just have:

```bash
geometry_namespace="$(ansi ${GEOMETRY_KUBE_NAMESPACE_COLOR:=white} ${geometry_namespace:=default})"
```

My reason for the "if / else" is because your terminal  foreground colour might not be white like an off-white or grey and the default white would overwrite this unwillingly.

This ways if you dont add `GEOMETRY_KUBE_NAMESPACE_COLOR` or `GEOMETRY_KUBE_CONTEXT_COLOR`, you can keep terminal profile colour settings. 

I guess this would be the proposed:

```bash
  if [[ ! -z "${GEOMETRY_KUBE_CONTEXT_COLOR}" ]]; then
    geometry_context="$(ansi ${GEOMETRY_KUBE_CONTEXT_COLOR} ${geometry_context})"
  fi

  if [[ ! -z "${GEOMETRY_KUBE_NAMESPACE_COLOR}" ]]; then
    geometry_namespace="$(ansi ${GEOMETRY_KUBE_NAMESPACE_COLOR} ${geometry_namespace:=default})"
  fi

# [...]

echo -n "${geometry_kube_separator}${geometry_kube} ${geometry_context}:${geometry_namespace:=default}${geometry_kube_separator}"
```

jedahan: I'd also be open to switching `ansi`'s argument order, and making the second argument optional. This would let us possibly remove all of the 'white' defaults easier.

jedahan: though that kinda messes with the ability to have text with spaces. Which ansi breaks right now too I guess?

jedahan: Okay, white can be changed to `default`

olegTarassov: yes ansi will break if empty.
the geometry_namespace needs the default because kubernetes namespace returns empty string sometimes when namespace is default. This behaviour is not seen observed on the context.

functions/geometry_kube.zsh
olegTarassov: let me try this out

olegTarassov:
Just Pushed the latest changes that are based on your recommendation.
I have one thing that I dont like or know how to fix in single line:

```bash
  kube_namespace="$(kubectl config view -o "jsonpath={.contexts[?(@.name==\"${kube_context}\")].context.namespace}" 2> /dev/null)"
  kube_namespace=${kube_namespace:=default}
```

I believe there might be a way to overwrite if empty to default

jedahan (approved):

jedahan:
Thanks for the updates, I rebased and merged in #266

PR #266, opened 2019-08-29
Feature/kube improvements

jedahan:
Thanks for all the hard work @olegTarassov!


jedahan (commented):
I rebased, but 90% of this work is based off of #264

PR #267, opened 2019-08-29
kube_namespace always resolves to default

olegTarassov:
Switch order because appending color to kube_context fails filter for kube_namespace

### DIFF
--- a/functions/geometry_kube.zsh
+++ b/functions/geometry_kube.zsh
@@ -18,7 +18,7 @@ geometry_kube() {
   kube_context="$(kubectl config current-context 2> /dev/null)"
   kube_namespace=$(kubectl config view -o "jsonpath={.contexts[?(@.name==\"${kube_context}\")].context.namespace}" 2> /dev/null)
 
-	kube_context=$(ansi ${GEOMETRY_KUBE_CONTEXT_COLOR:=default} ${kube_context})
+  kube_context=$(ansi ${GEOMETRY_KUBE_CONTEXT_COLOR:=default} ${kube_context})
   kube_namespace=$(ansi ${GEOMETRY_KUBE_NAMESPACE_COLOR:=default} ${kube_namespace:=default})
 
   if ( ${GEOMETRY_KUBE_VERSION:=true} ); then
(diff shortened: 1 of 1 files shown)


olegTarassov:
any update on this :)

kyounger:
This might be fixed with #271? It was just merged, so might be worth checking. With that PR, I get proper namespace showing. I did the refactor for that PR, so would be willing to investigate if your issue still persists.

olegTarassov:
Very nice refactoring!
Love the:

```bash
kubectl config view --minify --output "jsonpath={..namespace}
```

Yes it solves the issue but it removed the option to omit the k8s version print.

kyounger:
@olegTarassov like the git prompt function, you can now just piece together your own segments , and leave out the geometry_kube_version segment altogether. So it still allows you to omit it.

e.g. `GEOMETRY_RPROMPT=(geometry_git geometry_exec_time geometry_kube_symbol geometry_kube_context geometry_kube_namespace)`

olegTarassov:
that is definitely the better approach

olegTarassov:
just tested, it works but now you lose the separator because its not initialized:

```
  local separator=${GEOMETRY_KUBE_SEPARATOR:-"|"}
  echo -n ${(pj.$separator.)geometry_kube_details}
```

kyounger:
What's the value of your GEOMETRY_RPROMPT?

olegTarassov:
```
GEOMETRY_INFO=
GEOMETRY_GIT_GREP=ag
GEOMETRY_KUBE_CONTEXT_COLOR=red
GEOMETRY_KUBE_NAMESPACE_COLOR=cyan
GEOMETRY_KUBE_SEPARATOR="|"

GEOMETRY_RPROMPT+=(
    geometry_virtualenv
    geometry_git
    geometry_kube_symbol
    geometry_kube_context
    geometry_kube_namespace
)
```

kyounger:
Right the separator is only used with the `geometry_kube` function. If you piece your prompt together with the helper functions, you'll have to insert the separators yourself.

PR #271, opened 2019-09-14
Refactor kube

kyounger:
Use better approach for retreiving current namespace that doesn't rely on
getting context first.

Use the geometry_kube_{namespace, symbol, context, version} convention to
make this function similar to the git function.

Detect a valid current context in addition to a non-zero KUBECONFIG.

Add GEOMETRY_KUBE_VERSION_COLOR.
Remove GEOMETRY_KUBE_VERSION check, as you can remove this element now
with by intentionally excluding geometry_kube_version from your prompt.

Added some different logic to pinning. If you have have GEOMETRY_KUBE_PIN
unset, then it defaults to false and displays prompt extras only if
valid KUBECONFIG or kubectl get current-context is valid. If explictly
set to false, then it will not display at all. If set to true, then it
displays all the time regardless of a valid context.

### DIFF
--- a/functions/geometry_kube.zsh
+++ b/functions/geometry_kube.zsh
@@ -1,31 +1,38 @@
 # geometry_kube - show kubectl client version and current context/namespace.
+geometry_kube_symbol() {
+  ansi ${GEOMETRY_KUBE_COLOR:=blue} ${GEOMETRY_KUBE_SYMBOL:="⎈"}
+}
+
+geometry_kube_namespace() {
+  local kube_namespace=$(kubectl config view --minify --output "jsonpath={..namespace}" 2> /dev/null)
+  ansi ${GEOMETRY_KUBE_NAMESPACE_COLOR:=default} ${kube_namespace:=default}
+}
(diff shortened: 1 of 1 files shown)


kyounger:
The logic on pinning might not be acceptable, but wanted to put it out there. PR can always be updated to reflect maintainers' wishes it this regard.

jedahan (approved):
This looks fantastic, thank you for your consistent contributions

jedahan:
I think the pinning logic seems a little strange to me because it’s actually now three states - true, false, and null. But if it is useful in your case I’m happy to keep it until it becomes confusing for others

kyounger:
Yeah, I'm not married to the terminology.

It might be better to entirely replace with something more explicit like GEOMETRY_KUBE_MODE={always,never,smart}, and just default to 'smart'.

For a perspective on where I really want to take this: https://www.reddit.com/r/zsh/comments/d49xnl/modify_prompt_based_on_contents_of_current_buffer/

PR #273, opened 2019-10-11
feat: function lazy loading.

alxbl:
As mentioned in #272, this PR adds lazy loading capabilities to the function arrays. It will only source the functions that are requested in one of the prompts. I've done a quick spot check and it doesn't seem to cause any issues.

Cheers,
Alex

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -12,9 +12,13 @@ GEOMETRY[ROOT]=${0:A:h}
 (($+GEOMETRY_RPROMPT)) || GEOMETRY_RPROMPT=(geometry_exec_time geometry_git geometry_hg geometry_echo)
 (($+GEOMETRY_INFO)) || GEOMETRY_INFO=(geometry_hostname geometry_jobs)
 
+load=($GEOMETRY_PROMPT)
+load+=(${GEOMETRY_RPROMPT:|load})
+load+=(${GEOMETRY_INFO:|load})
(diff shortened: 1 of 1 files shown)


jedahan:
So I think the only downside of this, is if someone adds a function to either array that was supposed to be sourced first.

I think by understanding autoload/fpath we could have our cake and eat it too, but we might have to ask #zsh-users for help.

jedahan:
I would love to have this feature, btw

alxbl:
My understanding is that each function is self contained. If the plan is to have some sort of dependency management between functions then I agree that this approach won't work. I will read the manual for autoload and see if something can be done.

alxbl:
Good news!

autoload works very nicely. It can even be used without touching `fpath` manually by specifying absolute paths. I've still got to test it with the existing plugin system, but initial tests show that 

`autoload -Uk /absolute/path/to/function_name` will have the desired behavior of running the initialization code once upon first call and calling the function. The only caveat right now is that function files are suffixed with `.zsh`, and I think that this won't play nice with autoload (which expects `file_name==function_name` for the auto-call to work. I'll see if this can be modified, otherwise it might be necessary to follow autoload's syntax.

Cheers,
Alex

jedahan:
Thanks for the research and testing. I think it’s fine to rename the function files to not have a zsh extension.

PR #274, opened 2019-10-14
Unhide the geometry separator

dio-az:
The `GEOMETRY_SEPARATOR` still being used in the following places:

- https://github.com/geometry-zsh/geometry/blob/mnml/geometry.zsh#L81
- https://github.com/geometry-zsh/geometry/blob/mnml/geometry.zsh#L61
- https://github.com/geometry-zsh/geometry/blob/mnml/readme.md#general

Since it's exposed to the user, hiding it may be a **BREAKING CHANGE**.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -6,7 +6,7 @@
 
 typeset -gA GEOMETRY
 GEOMETRY[ROOT]=${0:A:h}
-: ${GEOMETRY[SEPARATOR]:=" "}
+: ${GEOMETRY_SEPARATOR:=" "}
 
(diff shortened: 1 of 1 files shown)


jedahan:
Thanks for the PR! Since this branch is about to released as v2 (as we are making lots of breaking changes), how would you feel about the following:

* https://github.com/geometry-zsh/geometry/blob/mnml/geometry.zsh#L81
  switch `GEOMETRY_SEPARATOR` to a hardcoded space, as this always felt hacky anyway.

* https://github.com/geometry-zsh/geometry/blob/mnml/geometry.zsh#L61
  fix to be `${GEOMETRY_SEPARATOR:-' '}` since theres really no reason to export the `GEOMETRY[SEPARATOR]` in the first place?

PR #288, opened 2019-12-23
Make host based color a generic function.

crasx:
I love the idea of using a random color that is generated based on host. Currently this is possible in `geometry_status` by setting `GEOMETRY_STATUS_COLOR_HASH` = true

In this PR I move the hash logic into a global place so it can be used for any variable. 
Here is an example of it being used: https://github.com/crasx/zsh/blob/214a2414f40e691bd6bd378cf658b876742fcfe2/.zshrc#L42
```
GEOMETRY_PATH_COLOR=`geometry::hostcolor`
```
If this is something you are interested in, let me know how you would like me to document it and if any other PR updates are required.


jedahan:
This looks great! Two things that would be nice to see before merging:

- document GEOMETRY_HOST_COLOR, preferably a screenshot with a split 2 or 3 window setup of why its useful (I think its super cool)

- show an example in the readme how to use geometry::hostcolor with status or path

crasx:
Thanks for the notes! I updated the readme with more details. I think for consistency it might be better if you generate the screenshot. Let me know if you need further changes. I'm still trying to improve my documentation skills.

readme.md
```diff
@@ -113,9 +110,22 @@ GEOMETRY_GIT_TIME_DETAILED=true     # show full time (e.g. `12h 30m 53s`) instea
...
+
+```shell
+GEOMETRY_HOST_COLORS=({1..7})  # Only use the colors 1-7
+GEOMETRY_HOST_COLOR=4          # Override the color for a specific host. 
+``` 
+![colorize](/images/screenshots/colorize.png)
+
 ## Thanks
 
-geometry is maintained by [frm](https://github.com/frm), [desyncr](https://github.com/desyncr) and [jedahan](https://github.com/jedahan).
```
jedahan, line 118: its spelled lowercase (or SAMECASE) everywhere else would prefer to leave this untouched.

jedahan:
@chrisfleming heads up I saw your dotfiles set `GEOMETRY_STATUS_COLOR_HASH` so you will need to change this a bit

jedahan:
Thank you so much @crasx!

PR #293, opened 2020-01-31
Refactor geometry_git_symbol into geometry_git

Konsonanz:
#### Description
This PR refactors the `geometry_git_symbol()` function into the main `geometry_git()` function().

With the recent refactoring of printing function-output with `geometry::wrap()`, the array-joining in `geometry_git_symbol()` is now un-needed. The call to sub-functions should therefore be included in the main-function.
#### To-fix
The current behaviour of `geometry_git_symbol()` would always insert one un-needed space if only one of the called sub-functions is called. Using `geometry::wrap()` fixes that (please note the spacing around the "remote"-icon):
demo
I also took a look into the other functions, but found no oppurtunitys where this could also be done (I guess `geometry_git_symbol()` was just accidentally missed in the previous refactoring).

### DIFF
--- a/functions/geometry_git
+++ b/functions/geometry_git
@@ -50,8 +50,6 @@ geometry_git_remote() {
   echo "$unpushed $unpulled"
 }
 
-geometry_git_symbol() { echo ${(j: :):-$(geometry_git_rebase) $(geometry_git_remote)}; }
-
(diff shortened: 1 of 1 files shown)


jedahan:
Awesome, thank you!

PR #294, opened 2020-04-19
Fix git conflict counting with multiple files

gluxon:
When the `$conflicts` shell variable contains newlines, the `$_grep` command attempts to read the entire output as a single file. This results in "No such file or directory" errors.

The fix here uses `xargs` to pass each file name individually.

I was seeing this on macoS 10.15.4 with zsh 5.8 (x86_64-apple-darwin19.3.0).

Screen Shot 2020-04-19 at 3 50 16 PM

### DIFF
--- a/functions/geometry_git
+++ b/functions/geometry_git
@@ -60,7 +60,8 @@ geometry_git_conflicts() {
   [[ -z "$conflicts" ]] && return
 
   _grep=${GEOMETRY_GIT_GREP:=${commands[rg]:=${commands[ag]:=${commands[grep]}}}}
-  conflict_list=$($_grep -cH '^=======$' $conflicts)
+  # Replacing \n with \0 since the -0 xargs flag is more portable than -d '\n'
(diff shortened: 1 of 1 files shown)


alxbl (approved):
Nice! I had this issue too but was too lazy to investigate it! The changes look good to me, but I'll let @jedahan have the final say in merging.

gluxon:
@alxbl Thanks for the review. @jedahan Any chance this can get a look? Thanks!

jedahan:
This looks good, though I wonder if we can use zsh's env substitution to achieve the same thing. Something like 

`conflict_list=$($_grep -cH '^=======$' ${conflicts/\n/\0/})`

gluxon:
I think that should work as well and is indeed much simpler. Let me make sure later tonight and update this PR if there's no issues. Thanks!

gluxon:
Looks like we're going with #299, which has a more intuitive solution. 🙂 Closing this PR for now.

PR #297, opened 2020-06-12
Fix whitespace escape in git_status

coteeq:
In git_status, when path to git dir contains whitespace
`git ls-files` receives only part of that path and shows kinda annoying fatal error


jedahan:
Thank you!

PR #299, opened 2020-07-27
Fix git conflict counting with multiple files, thanks @gluxon

jedahan:
@gluxon came up with a solution to "no such file or directory"
errors, which exposed underlying issues with our git conflict counting.

The solution was to use null characters and xargs instead of newlines.

`git diff` has the `-z` flag to list names with \0 instead of \n,
and zsh has an expansion modifier `0` which expands on \0 instead of \n.

This also fixes the `20f` issue when only one file has conflicts, it
was counting characters of the filename instead of the file list.

Thank you @gluxon!


gluxon:
Ah this is much more elegant. Thank you!

PR #301, opened 2020-08-14
git_status: simplify tracked file dirty check

jedahan:
i was getting some weird quoting behaviour before in my shell, this seems reasonable

### DIFF
--- a/functions/geometry_git
+++ b/functions/geometry_git
@@ -28,8 +28,8 @@ geometry_git_branch() {
 geometry_git_status() {
   command git rev-parse --git-dir > /dev/null 2>&1 || return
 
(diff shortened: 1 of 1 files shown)


jedahan:
i am not sure if i should use HEAD or something else, as maybe if you are in detached head this breaks?

frm:
re the HEAD subject, I was having issues in detached HEAD a few versions back so that reasoning makes sense. Can you try and see what happens with no argument?

jedahan:
I tested in a detached head and it still seems to work. git diff requires a commit-ish thing, so HEAD seems fine.

no argument doesn't work.

frm (approved):
Cool, let's go for it

PR #302, opened 2020-09-20
Allow customization of terminal title

duncanbeevers:
### Summary
Previously, geometry showed output in three locations:
* `GEOMETRY_PROMPT`
* `GEOMETRY_RPROMPT`
* `GEOMETRY_INFO`

This PR adds two new display locations:
* `GEOMETRY_TITLE`
* `GEOMETRY_RUNTITLE`

### Details

Geometry was already implicitly updating the terminal title.
This PR allows configuration of these locations through the same paradigm as drives the rest of geometry.

The original title-controlling functions; `geometry::clear_title`, and `geometry::set_title` ran with an implicit lifecycle.

#### De-coloring

The biggest challenge with this PR was de-coloring the output of geometry plugins.
The `deansi` function included here is hand-rolled and worked with the handful of functions I tried out; including`geometry_path`,`geometry_node`, and `geometry_hostname`.

#### Limited testing

I've primarily developed/tested this in iTerm. It looks a little funny in Terminal.app
iTerm | Terminal.app
-|-
2020-09-20 at 12 47 PM | 2020-09-20 at 12 43 PM

#### No current command

This PR is a regression, since the currently-executing command is no longer exposed in the terminal title.
It seems simplest to expose it as a generic geometry function, like `geometry_cmd` or something, but it may only be useful when used in `GEOMETRY_RUNTITLE`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Add GEOMETRY_TITLE and GEOMETRY_CMDTITLE as display locations
+- Add geometry_cmd function to display currently-running command in GEOMETRY_CMDTITLE
 
 ### Fixed
 - Fix git functions erroring out in non-git directories (thanks @duncanbeevers!)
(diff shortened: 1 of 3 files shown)


jedahan:
Great idea! I'd be happy to merge with a few minor changes:

- Add this feature to the changelog
- Set the default runtitle to show the currently running command (with or without `geometry_cmd` as an external function, I am not sure it makes sense externally)

- (optional) Think if there could be a better name for RUNTITLE, where better could mean more consistent with zsh naming of things. For example, CMDTITLE since it is precmd. I am happy to merge as RUNTITLE though.

duncanbeevers:
@jedahan I pushed a change that adds a `geometry_cmd` internal function used for displaying the currently-executing / last-executed command.

I didn't find a nice way to restore the `" @ "` in between the command and PWD, like in the old baked-in term title.
I thought maybe there was a way to do it with `geometry_echo`.

Any suggestions here would be appreciated.

2020-09-25 at 9 24 AM

jedahan:
I am happy keeping it as is (no " @ ") - geometry::wrap uses GEOMETRY_SEPARATOR, which is more consistent with the other output

duncanbeevers:
@jedahan I've squashed this down to a single commit and updated the commit message with a summary and explanation of the changes.

jedahan:
Thanks a bunch @duncanbeevers !

PR #307, opened 2021-02-08
use oh-my-zsh title() function to support more terminals

itsxaos:
Setting title was broken for Alacritty and xfce4-terminal, so I took the `title()` function from oh-my-zsh (see [here](https://github.com/ohmyzsh/ohmyzsh/blob/master/lib/termsupport.zsh))


frm:
I'll get around to test this later, but can you include a comment linking the original function? 🙏

PR #308, opened 2021-02-08
Allow customizing path truncation

itsxaos:
Replaced hard-coded 3 with `GEOMETRY_PATH_TRUNCATE` variable (defaults to 3).
Now maximum amount of directories to show can be customized by setting that variable.

This is implemented a little different to how the other defaults are implemented, but it seemed more efficient to just set the variable to the default if it wasn't set, because the variable needs to be checked anyway.

### DIFF
--- a/functions/geometry_path
+++ b/functions/geometry_path
@@ -1,7 +1,7 @@
 # geometry_path - show the current path
 
 geometry_path() {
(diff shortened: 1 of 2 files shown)


frm (approved):
Thanks for this! LGTM

PR #309, opened 2021-04-25
Silence warnings about line endings

MageJohn:
I was opening a repo that had a mix of line-endings, and I got warnings about line ending changes on every prompt; since they aren't useful when repeated at every prompt, I identified the command that was producing them and directed it's stderr to `/dev/null`.

### DIFF
--- a/functions/geometry_git
+++ b/functions/geometry_git
@@ -65,7 +65,7 @@ geometry_git_conflicts() {
   local conflicts conflict_list
   local file_count
   local total raw_total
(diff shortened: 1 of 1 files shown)


frm (approved):
LGTM

MageJohn:
@frm sounds like a plan 🥳 I don't have write access though, so someone who does will need to press the button!

frm:
My bad 👍 thanks for the ping

PR #317, opened 2023-12-19
Compatibility with znap

cweagans:
Geometry currently doesn't work with [znap](https://github.com/marlonrichert/zsh-snap). Details here: https://github.com/marlonrichert/zsh-snap/issues/67

This PR fixes that problem without changing any Geometry behaviors + documents how to install with znap.

### DIFF
--- a/geometry.zsh
+++ b/geometry.zsh
@@ -15,7 +15,7 @@ typeset -gA GEOMETRY; GEOMETRY[ROOT]=${0:A:h}
 
 autoload -U add-zsh-hook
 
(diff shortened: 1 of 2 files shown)


jedahan:
The functions directory is part of the source code of geometry, it should be read-only. Updating now, and making our sourcing better in `use-fpath` branch. Is there a minimal test reproduction we can try to make sure this is compatible with znap?

jedahan:
Trying another stab it at with the `function-safelist` branch

PR #320, opened 2025-01-02
Print geometry_info in statusline

jedahan:
This is a significant change, as we dont just run a command as if the user typed it. Requires ^L to clear, but seems less overall opportunity for bugs.


jedahan:
manually added in another commit

PR #327, opened 2025-07-06
prefix git conflict grep path with git repo root

### LINKED ISSUE #326: Merge conflict display fails if CWD is not git repo root
https://github.com/geometry-zsh/geometry/blob/0f82c567db277024f340b5854a646094d194a31f/functions/geometry_git#L62-L85

when the list of conflicts is created with `git diff`, all paths are relative from the repo root. then, the grep call on line 73 searching for the conflict symbol will grep from the CWD, which is not necessarily the repo root. this will cause the search to fail, and possibly display an error for every file that couldn't be pathed.

a simple fix would be to prefix each item in the list with `git rev-parse --show-toplevel` which should return the git root.

more ideally, it might be preferable to use `git diff --check`, which contains a built-in tool for detecting merge conflicts.

turbocrime:
fixes https://github.com/geometry-zsh/geometry/issues/326


functions/geometry_git
```diff
@@ -65,12 +65,17 @@ geometry_git_conflicts() {
...
 
   [[ -z "$conflicts" ]] && return
 
+
   _grep=${GEOMETRY_GIT_GREP:=${commands[rg]:=${commands[ag]:=${commands[grep]}}}}
-  conflict_list=$($_grep -cH '^=======$' ${(0)conflicts})
+
+  git_root=$(git rev-parse --show-toplevel)
+
+  conflict_list=$($_grep -cH '^=======$' ${git_root/$^conflicts})
```
turbocrime, line 78: this expansion feels fragile, not sure of best practice here

